PR #4, opened 2017-11-25
Update README.md

Nightcaper:
Having to change 'xit's in exercises like repeatString for testing exercise solutions was not immediately obvious. I felt it should be mentioned so newbies before me won't face confusing jasmine messages about xits when testing their code.

I don't know how the wording of the readme is really intended, so I hope my change will address this properly.


codyloyd:
gah!  Good catch!  This is the change that i had made that we lost when I undid my stupid 'solutions' merge.

Thanks for this PR 👍

PR #8, opened 2018-01-01
zero factorial is 1

Dareon4:
The zero factorial is equal to 1.
[Factorial - Wikipedia](https://en.wikipedia.org/wiki/Factorial#Definition)


codyloyd:
nice catch!!  thanks!

Dareon4:
You're welcome!

PR #9, opened 2018-01-03
Corrected solution to factorial function

ThirtyThreeB:


codyloyd:
decided to keep the non-recursive one as well... just so we don't break other student's minds ;)

Thanks for this!

PR #19, opened 2018-06-09
Rewording instructions

EBoisseauSierra:
While doing the suggested exercises, I've improved the wording of two of them (`repeatString` and `tempConv`), as well as improved "correctness" (use `let` instead of `var`, capitalize titles).

### DIFF
--- a/caesar/caesar.js
+++ b/caesar/caesar.js
@@ -1,4 +1,4 @@
-var caesar = function() {
+let caesar = function() {
 
 }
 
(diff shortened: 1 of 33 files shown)


codyloyd:
if we're going to be pedantic about "correctness" all the `let`s should be `const`

the other changes look good though

EBoisseauSierra:
indeed, thanks for the hint. should be better now.

codyloyd:
🎉 thanks

PR #28, opened 2018-12-03
Update installation instructions

leila-alderman:
Added in a step for installing Node and clarified to not follow Linux's suggestions.


leosoaivan:
LGTM

PR #68, opened 2020-03-28
Fundamentals Part 4

AndresGarcia98:
Here there are the solutions of the points needed


I3uckwheat:
These are not supposed to be filled out, that way others can do the lesson as well.

PR #73, opened 2020-05-01
Added random test

Timework:
I added a random test getting a number from 1 to 1000. You said you wanted comments on it but I was not sure where to put them. I added the global modifier to the recommended regex in order for it count all the words properly.


repeatString/repeatString.spec.js
```diff
@@ -16,4 +16,8 @@ describe('repeatString', function() {
   xit('returns ERROR with negative numbers', function() {
     expect(repeatString('hey', -1)).toEqual('ERROR');
   });
+  xit('repeats the string a random amount of times', function () {
+    let number = Math.floor(Math.random() * 1000)
```
I3uckwheat, line 20: This let, should be a `const`. There should also be a comment here letting the reader know how this works.

I3uckwheat:
@Timework Great, looks amazing!

PR #111, opened 2020-10-03
calculator done

bmukha:


I3uckwheat:
Good job! But we don't want to merge this. We would like others to be able to do the same exercise as you did, and merging this would mean they just have the answers! That's no good! I'm closing this so others don't find it and cheat.

PR #112, opened 2020-10-08
Update README.md

Sh1710:


README.md
```diff
@@ -7,6 +8,7 @@ There will eventually be a suggested order of completion, but at this time since
 1. Reverse String
 
 ## HOW TO USE THESE EXERCISES
+## READ THIS CAREFULLY
```
I3uckwheat, line 11: image


When using two headers, the markdown is not displayed properly

I3uckwheat (changes requested):

README.md
I3uckwheat: Instead of using two headers, one can just make a new **BOLD** paragraph under the header

PR #113, opened 2020-10-10
sumAll solution: check if min/max are integers, not just numbers

mangonaise:


I3uckwheat:
This is very early in the curriculum, I don't think we want to confuse people on integers vs numbers. The tests are only passing in numbers, A better solution would be to change "integers" to "numbers" in the read-me for that lesson, as well as add some non-integer tests in the spec file.

Good catch though! Please make another PR with those changes :)

mangonaise:
@I3uckwheat I think I disagree with your alternative solution - I can't see how sumAll() makes sense with non-integer values.

I3uckwheat:
`1.2 + 1.4 + 0.3`

These numbers can be summed to be: `2.9` which is not an integer. So: `sumAll([1.2, 1.4, 0.3])` can be perfectly valid.

I3uckwheat:
@mangonaise I would be interested to know why you think it doesn't make sense however. In case I'm missing your point.

mangonaise:
@I3uckwheat 

Appreciate the open-minded reply. According to the exercise description, sumAll() takes 2 integers and sums all the values between them; it doesn't take an array of numbers and sum them, as you suggested. This is the task:

> Implement a function that takes 2 integers and returns the sum of every number between(and including) them:
> 
> `sumAll(1, 4) // returns the sum of 1 + 2 + 3 + 4 which is 10`

So maybe actually the exercise itself is poorly named.

I3uckwheat:
@mangonaise 

Which is why I suggested:

>  A better solution would be to change "integers" to "numbers" in the read-me for that lesson, as well as add some non-integer tests in the spec file.

I3uckwheat:
@mangonaise Ohhh, I *did* miss something. I agree that the lesson is poorly named.

I3uckwheat:
I will merge this, I agree with your solution, even though it may be a bit more advanced than what we want people doing at that stage. It may get them to look at the documentation a bit.

I3uckwheat:
@mangonaise 

Would you care to update the name of this lesson? Maybe "addUpTo"?

mangonaise:
@I3uckwheat 

Sure, I'll think of a new name. Not a fan of "addUpTo", it's more like "sumBetween" or something. I also thought some other exercises could benefit from more descriptive names so I might look at those. Should I do multiple commits or combine the changes into a single commit?

I3uckwheat:
@mangonaise multiple commits are easier for us to review

PR #116, opened 2020-10-17
Update readme.md

Abhineet0011:
Changed spelling mistakes


xandora:
The first edit is fine, although there should not be a space between `'The Odin Project'` and the full stop.

The second change does not make sense. Why put a comma after `so`? There is not a natural pause there when reading that sentence.

xandora:
Whoops, wrong button... Where's the request change button gone?! Anyway, if you adjust the PR we can accept it.

marvingay:
I made additional revisions to the README so it reads a tad bit better.

@Abhineet0011  Thank you for your contribution.

xandora (commented):
looks good.

PR #120, opened 2020-12-22
Update helloWorld.js

vrarora:
I added the answer instead of empty strings as solution must be added in the solutions branch.


xandora:
It's an easy enough exercise, people shouldn't need to look for the solution to this one. However, we do love consistency... So looks good to me.

PR #123, opened 2021-01-12
Updated README

GroundedCastle:
There were a lot of issues related to capitalization, syntax, and cogency.


README.md
```diff
@@ -1,25 +1,25 @@
...
+There will eventually be a suggested order of completion, but at this time since we are still in the process of creating more exercises the order is subject to change and has not yet been specified. However, there are a few exercises that make a good "starting point". Feel free to at least start with these:
 
 1. Hello World
-1. Repeat String
-1. Reverse String
+2. Repeat String
+3. Reverse String
 
-## HOW TO USE THESE EXERCISES
+## How to Use These Exercises
```
twalton83, line 9: If we are going to change the case here, we may as well just capitalize "To" as well.

README.md
```diff
@@ -1,25 +1,25 @@
...
-3. Jasmine.  Jasmine is a testing framework for Javascript.  Type `jasmine -v` to check for it.  If you need to install it, type `npm install -g jasmine` to do so.
-4. Clone this repo and get started.
+1. NPM.  To check if you have NPM installed, type `npm --version` in your terminal. If you get back `Command 'npm' not found, but can be installed with:`, do NOT follow the instructions in the terminal to install with `apt-get`. (This causes permission issues.) Instead, install Node with NVM by following the instructions [here](https://github.com/TheOdinProject/curriculum/blob/master/foundations/installations/installing_node.md).
+2. Jasmine.  Jasmine is a testing framework for JavaScript. Type `jasmine -v` to check for it. If you need to install it, type `npm install -g jasmine` to do so.
+3. A copy of this repository. Copies of repositories on your machine are called clones. If you need help cloning, you can learn how [here](https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/duplicating-a-repository)
```
twalton83, line 13: At this point, the students should already be familiar with cloning due to other prior lessons.

README.md
```diff
@@ -1,25 +1,25 @@
...
+Each exercise includes 3 files: a markdown file with a description of the task, an empty (or mostly empty) JavaScript file, and a set of tests.  To complete an exercise, you'll need to go to the exercise directory with `cd exerciseName` in the terminal and run `jasmine exerciseName.spec.js`.  This should run the test file and show you the output.  When you first run a test, it will fail.  This is by design! You must open `cd exerciseName` and write the code needed to get the test to pass. Some of the exercises have test conditions defined in their spec file that are defined as 'xit' compared to 'it'. This is purposeful. After you pass you first 'it', you will change the next 'xit' to an 'it' and test your code again. You'll do this until all conditions are satisfied.
```
twalton83: Small thing but there's an extra space before "  This is by design!". 

Also, can we sneak in calling the "xit" and "it"s "test cases" at some point?

README.md
```diff
@@ -1,25 +1,25 @@
...
 
-The first exercise, `helloWorld` will walk you through the process in more depth.
+The first exercise, `helloWorld`, will help the process become clearer.
```
twalton83, line 17: I don't think this change adds anything that the prior iteration of the sentence didn't have. I'd rather we change this to "The first exercise, `helloWorld`, will walk you through the process in-depth."

twalton83 (changes requested):
Made a few quick suggestions.

README.md
```diff
@@ -1,25 +1,25 @@
...
+Each exercise includes 3 files: a markdown file with a description of the task, an empty (or mostly empty) JavaScript file, and a set of tests.  To complete an exercise, you'll need to go to the exercise directory with `cd exerciseName` in the terminal and run `jasmine exerciseName.spec.js`.  This should run the test file and show you the output.  When you first run a test, it will fail.  This is by design! You must open `cd exerciseName` and write the code needed to get the test to pass. Some of the exercises have test conditions defined in their spec file that are defined as 'xit' compared to 'it'. This is purposeful. After you pass your first 'it', you will change the next 'xit' to an 'it' and test your code again. You'll do this until all conditions are satisfied.
```
twalton83, line 15: "You must open `cd exerciseName`" at this point, they are already `cd`'ed into the directory, and you can't `cd` into a file, you need to open the file, which at that point is already in their VSC sidebar, so we can just say "open the exercise file" or something to that effect.

twalton83 (changes requested):

README.md
xandora: There are quite a few double spaces after periods. We're not using typewriters here, let's stick to a single space after a period.

README.md
GroundedCastle: I am going through Foundations, and there's almost no practice with cloning. There was likely brief exposure to it, but not enough to be able to copy the repository without first searching on Google.

README.md
GroundedCastle: You're right that exerciseName shouldn't be said. I meant to say that you need to open the file that isn't the spec file. Do you think that's needed for clarification?

README.md
twalton83: We have multiple lessons where one indeed does clone repositories- that said, after some thought, I don't see the need to remove this so we can keep it, given that people can potentially contribute at any stage in their learning 😄

README.md
twalton83: I think we can clarify that, let's just use a broader "open the file" or use the correct command here.

twalton83 (approved):
Looks good to me, thank you!

PR #137, opened 2021-04-27
Optimize regex in palindrome.js solution

MrBlamz:
Since we use toLowerCase() before using replace there is no need to use A-Z in our regex cause it will be impossible to have uppercase characters in the string.


kemonprogrammer:
I had the same idea and it works too, why was this pull request closed?

PR #142, opened 2021-05-06
Correct exercise number

kemonprogrammer:
Because `repeatString/` is Exercise 02
and `removeFromArray/` is Exercise 04


twalton83 (approved):
Thanks!

PR #143, opened 2021-05-06
Change function names to camelCase

kemonprogrammer:


twalton83:
This doesn't exactly make sense with these particular functions, given that they refer to acronyms. A similar example would be `atob` and `btoa`: https://stackoverflow.com/questions/33854103/why-were-javascript-atob-and-btoa-named-like-that#:~:text=atob%20stands%20for%20ASCII%20to%20binary

PR #149, opened 2021-05-13
Implement Jest testing platform for all exercises

xandora:
This PR overhauls the JavaScript exercises to use Jest instead of Jasmine.

- Updated README.md content.
- Updated syntax to be consistent across exercises.
- Users can simply npm install to pull required node modules in using package.json file.
- ESLint has been included, but is set to run manually as a forward thinking possibility for a "Linting lesson" using existing lessons.


.eslintignore
```diff
@@ -0,0 +1 @@
+generator-exercise/
```
twalton83, line 1: Newline me!

.gitignore
```diff
@@ -1,4 +1,3 @@
 .vscode
-node_modules
-package-lock.json
-package.json
\ No newline at end of file
+node_modules/
+.DS_Store
```
twalton83, line 3: One more here!

twalton83 (commented):
Once those new lines are fixed up, we are good to go!

PR #156, opened 2021-05-28
helloWorld.spec.js: Fix disparity between the README example and spec file

xandora:
Quick fix to bring the actual spec file in line with the example in the README.


KevinMulhern (approved):
LGTM 👍

PR #164, opened 2021-07-20
update README.md

mc811mc:
I just went through this problem and ended up debugging by looking at the sumAll.spec.js log on my console because the README.md did not list the exceptions the function should have considered. I believe it would be more beneficial to provide students what they should be considering when implementing the following question.


twalton83:
The students are directed to read the test cases to see what they should consider.

PR #170, opened 2021-08-07
Order exercise folders

### LINKED ISSUE #169: Add numbering on exercise folders
It would be more intuitive if the folders were numbered so that they can be done in order. For example, they could be called "01 Hello World", "02 Repeat String", and so on. Their number was mentioned in each of their individual README files, but having the folder themselves labelled doesn't add much effort and makes things easier to go through.

I know the project README says that since it's "still in the process of creating more exercises the order is subject to change and has not yet been specified", but The Odin Project courses already recommend a specific order so I think it can be added on the folder name

remtaine:
Ordered the exercise folders by placing their numbers in their folder name

Closes #169


twalton83 (changes requested):
Generally we want to avoid spaces in directories, we need to use underscores.

twalton83:
Thank you for doing this. A similar PR needs to be made for the solutions branch- you may be able to make a PR using the same branch, instead targeting the solutions branch here.

twalton83 (approved):

PR #190, opened 2021-11-04
Add a test case for the multiply function

dragi-ns:
My original solution would incorrectly return `1` for an empty array, but it passes all the current tests.
```javascript
const multiply = function(operands) {
  return operands.reduce((total, operand) => total *= operand, 1);
};
```
New test checks that multiply function returns `0` for an empty array.

### DIFF
--- a/08_calculator/calculator.spec.js
+++ b/08_calculator/calculator.spec.js
@@ -39,6 +39,9 @@ describe('sum', () => {
 });
 
 describe('multiply', () => {
+	test.skip('multiplies an empty array', () => {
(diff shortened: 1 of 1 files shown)


xandora:
I feel like if you run a function like this without any input, it shouldn't give you a valid result.

dragi-ns:
But, when given an empty array, it should return `0`, like the `sum` function.

https://github.com/TheOdinProject/javascript-exercises/blob/6b0e0e55e36e7cb1f2deb499e16389c6ccf778f3/08_calculator/calculator.spec.js#L25

PR #191, opened 2021-11-04
Add new test cases for palindromic numbers

dragi-ns:
New tests check for numeric palindromes.


twalton83:
I don’t think we want to add more test cases here.

PR #206, opened 2021-12-27
Remove predefined parameter from reverseString.js

chagstep1:
Making it consistent with other exercise setups


xandora (approved):
Consistency is nice. Good work.

PR #211, opened 2022-01-05
Add test for empty strings

gerijeb:
Pass empty string as part of the arguments to test.


05_sumAll/sumAll.spec.js
```diff
@@ -19,4 +19,7 @@ describe('sumAll', () => {
   test.skip('returns ERROR with non-number parameters', () => {
     expect(sumAll(10, [90, 1])).toEqual('ERROR');
   });
+  test.skip('returns ERROR with empty strings', () => {
```
twalton83, line 22: 'returns ERROR with non-number parameters' covers this test case.

PR #214, opened 2022-01-17
Update helloWorld.js

Joaoblink:


CouchofTomato:
Thanks @Joaoblink but we can't merge solutions here as we need it to be blank for future students

PR #216, opened 2022-01-23
Fixed semicolons

icorvus:
I found some missing semicolons and decided to fix them


xandora:
Yep, we probably should use semicolons considering they're *technically* not optional.

PR #218, opened 2022-01-26
added alternate solution to multiply using Exponentiation operator

trekitch:
Added a small alternative solution to one of the problems in the calculator project.

### DIFF
--- a/calculator/calculator.js
+++ b/calculator/calculator.js
@@ -16,7 +16,7 @@ const multiply = function(array) {
       : 0;
 };
 
-const power = function(a, b) {
+const altPower = function(a, b) {
   return Math.pow(a, b);
 };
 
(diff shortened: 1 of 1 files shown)


xandora:
The first problem I see here is that this code *should* still run. Will it run with two functions with the same name?

trekitch:
@xandora I was under the impression that the Jest solutions branch was just used to show the solutions. I got the idea for the alternate solution from the recursive factorial alternate solution. I do see that that one is name different but still not used. I have update my alternate function to reflect that.

calculator/calculator.js
```diff
@@ -20,6 +20,11 @@ const power = function(a, b) {
   return Math.pow(a, b);
 };
 
+//alternate solution using Exponentiation opertator
```
xandora, line 23:
```suggestion
//alternate solution using the Exponentiation operator
```

xandora (changes requested):
Just one small change required, but this looks good to me!

trekitch:
Accepted changes

PR #219, opened 2022-01-28
Improve the readability of the RegExp

tunsmart:
Hi,

While both the original code and the new code do the same thing, I feel like the new one reads better especially for beginners just getting a hang of RegExp in Javascript.

### DIFF
--- a/palindromes/palindromes.js
+++ b/palindromes/palindromes.js
@@ -1,5 +1,5 @@
 const palindromes = function(string) {
-  processedString = string.toLowerCase().replace(/[^A-Za-z]/g, "");
+  processedString = string.toLowerCase().replace(/[^a-z]/g, "");
   return (
     processedString
       .split("")


twalton83:
This actually doesn't account for uppercase.

Screen Shot 2022-01-28 at 4 30 15 PM
Screen Shot 2022-01-28 at 4 29 59 PM

tunsmart:
Thanks for your response. That's very true. But we don't need the RegExp to account for uppercase again since `.toLowerCase()` is called before the RegExp. I guess I should have qualified my previous statement that both RegExps are essentially doing the same thing _in this specific case_. And in any case, all the tests, including the one specifically testing for uppercase, passed with the new code too.

tunsmart:
But maybe this is a non-issue anyways, but I just thought I should this chip in

twalton83:
You're right, facepalmed at that. Thank you.

PR #222, opened 2022-02-01
Fix the for loop

Morthiner:
Changing the for loop's condition so that it can fit the tests


Mclilzee:
In order for this to work you have to remove the `+ 1` next to max. otherwise this will actually break the for loop, which is working properly.

PR #223, opened 2022-02-07
08 calculator: add some unit tests.

helloShen:
Add some unit test for `power()` function to make the test more robust.
1. tests with negative base.
2. tests for the negative exponent.
3. if exponent equals `0`, result should always be `1`.
    3.1. except for a base of `0`, which should return `0`.
4. if base equals `0`, 
    4.1. return `Infinity` if exponent is negative.
    4.2. otherwise, always return `0`.

By the way,  a test for an empty array input is added for `multiply()` function.

### DIFF
--- a/08_calculator/calculator.spec.js
+++ b/08_calculator/calculator.spec.js
@@ -1,7 +1,7 @@
 const calculator = require('./calculator');
 
 describe('add', () => {
-	test.skip('adds 0 and 0', () => {
+	test('adds 0 and 0', () => {
 		expect(calculator.add(0,0)).toBe(0);
 	});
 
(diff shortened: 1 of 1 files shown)


helloShen:
Let me check it.

helloShen:
27 passed, 1 failed.
```js
expect(calculator.factorial(-1)).toBeUndefined();
```

Solution doesn't check whether the input is negative or not.
```js
const factorial = function(n) {
  if (n === 0) return 1;
  let product = 1;
  for (let i = n; i > 0; i--) {
    product *= i;
  }
  return product;
};
```
So as the recursive version,
```js
const recursiveFactorial = function(n) {
  if (n === 0) {
    return 1;
  }
  return n * recursiveFactorial (n-1);
};
```

helloShen:
If it's ok I can create another pull request to also update `calculator.js` and `calculator.spec.js` in solution branch.

PR #234, opened 2022-02-20
Exercises: Add solution directories for all exercises

thatblindgeye:
Rather than containing exercise solutions on a separate branch, they should follow the CSS Exercise repo and have a directory within each exercise. This makes it easier for users to compare their solutions with TOP's solutions upon completion of an exercise, and reduces some upkeep should a new exercise be added or a current exercise need updating (not needing to make two PRs for separate branches).

Due to there being open PRs regarding exercises, this should not be merged until those open PRs are merged and this PR is updated.


ChasingSingularity:
Why is this Not moving Forward?

xandora:
@ChasingSingularity 

> Due to there being open PRs regarding exercises, this should not be merged until those open PRs are merged and this PR is updated.

Probably the fact that the outstanding PRs mentioned in the original post were never resolved. Or OP forgot.

02_repeatString/solution/repeatString-solution.spec.js
```diff
@@ -0,0 +1,39 @@
+const repeatString = require("./repeatString");
+
+describe("repeatString", () => {
+  test("repeats the string", () => {
+    expect(repeatString("hey", 3)).toEqual("heyheyhey");
+  });
+  test.skip("repeats the string many times", () => {
```
richfromm, line 7: As asked at https://github.com/TheOdinProject/javascript-exercises/pull/325#issuecomment-1411208821, wouldn't it be better of the solutions all had `test()`, with no `test.skip()`, so one could easily verify that the solutions actually work?

02_repeatString/solution/repeatString-solution.spec.js
```diff
@@ -0,0 +1,39 @@
+const repeatString = require("./repeatString");
```
richfromm, line 1: If you want the solutions to be easily verifiable (see [below](https://github.com/TheOdinProject/javascript-exercises/pull/234/files#r1092794141), would it be better if each `foo-solution.spec.js` file require'd the corresponding `"./foo-solution"` rather than just `"./foo"`?

thatblindgeye: Good catch! Didn't realize I hadn't pushed that to my branch, should be up to date now!

02_repeatString/solution/repeatString-solution.spec.js
thatblindgeye: That's a really good point, thank you for the suggestion! 🎉

01_helloWorld/solution/helloWorld-solution.spec.js
```diff
@@ -1,7 +1,7 @@
-const helloWorld = require("./helloWorld");
+const helloWorld = require('./helloWorld-solution');
```
richfromm, line 1: Are all of the double quote to single quote changes just in the interest of consistency?

Does TOP have a style guide or some other policy that prefers single quotes?

xandora: This repository is set up to use eslint with the airbnb style guide, but it's probably been a while since anyone actually ran this code through a linter.

xandora: Unless... @thatblindgeye did you actually run eslint over this? :D

thatblindgeye: @xandora all I did was save the file 😆

xandora: I had a quick search through the main repository and curriculum repositories, and we don't explicitly mention that airbnb is our preferred style guide, but it is implied as part of the eslint config across the repositories.

* Contributions to the curriculum should follow the style guide mentioned in the [contributing guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md#top-curriculum-repo). 
* The [guide for contributions to the main site](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md#top-website-repo) doesn't specifically say "use airbnb", but we do have tests that must be passed which cover linting, so you don't need to think about it:
```
rubocop
yarn eslint
rspec
yarn test
```

We have no overall recommended list of extensions, except those that are mentioned during the curriculum at the appropriate time they need to be. There's no point in installing Prettier or Eslint in VSCode before we get to those lessons, as they will start clobbering a learners code in way they might not expect to see.

richfromm: Oh, didn't realize you get to that later in the curriculum. (I'm still going through it, approaching the end of the Foundations Course.) Sounds good, I can appreciate not wanting to overwhelm beginners.

thatblindgeye:
@01zulfi would you mind taking a look at this? Neither branch has had an update since this was most recently updated so there shouldn't be anything missing, probably just mainly pulling in my branch locally to make sure tests pass/fail as expected.

01zulfi:
@thatblindgeye Definitely, give me a couple days.

01zulfi (approved):
Pulled in the branch and everything works as expected. Tests fail/skip for incomplete functions, and tests pass for all the files in `**/solutions/`. 

This should be good to go :rocket:

PR #242, opened 2022-03-30
05_sumAll: Add new test

jamhog:
#113 altered the solution to check min/max are integers (not just numbers) but no test was added that performs this check.

This PR adds such a test.


thatblindgeye (changes requested):
Thanks for the update! Can you just confirm that this new test passes with a screenshot?

jamhog:
Created a new PR #304 that updates solutions and shows this passing.

thatblindgeye:
Closing this as it has been resolved by #304

PR #244, opened 2022-04-22
04_removeFromArray: New solution proposed

ayer21:
this solution is based on two nested loops, one that takes an item from the first array, and one that checks if it's equal to a value in the second array, by looping through the second array.
it also uses the splice method to remove the item from the array. so no need to create a new one.

-   [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)

<hr>

**1. Because:**

Closes #XXXXX

**2. This PR:**

### DIFF
--- a/removeFromArray/removeFromArray.js
+++ b/removeFromArray/removeFromArray.js
@@ -28,4 +28,18 @@ const removeFromArray = function (...args) {
 // }
 //
 
+// another solution is to loop through the first array items, and for each item, you check if it's on the second array. if true,
+// the item is removed from the array using splice() method.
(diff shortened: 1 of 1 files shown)


thatblindgeye (commented):
I'm debating whether it would be worth including a third possible solution for this. Is there a significant benefit to this proposed solution, or is it more just offering a third option?

thatblindgeye:
For now I will close this, as I'm not sure if there's a significant benefit to including yet another solution option.

PR #253, opened 2022-05-28
JS exercises solutions - removeFromArray: Make `array` an independent parameter

yossirise:
This spares us from having to extract `array` from `args` and is more idiomatic (I think).

-   [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)

<hr>

**1. Because:**

- A concise, idiomatic way to handle `array` would be to declare it as an independent parameter rather than extract it from the argument list

**2. This PR:**

- Moves array to the parameter list and removes code made redundant

### DIFF
--- a/removeFromArray/removeFromArray.js
+++ b/removeFromArray/removeFromArray.js
@@ -1,9 +1,7 @@
 // we have 2 solutions here, an easier one and a more advanced one.
 // The easiest way to get an array of all of the arguments that are passed to a function
 // is using the rest operator. If this is unfamiliar to you look it up!
-const removeFromArray = function (...args) {
-  // the very first item in our list of arguments is the array, we pull it out with args[0]
-  const array = args[0];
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):

removeFromArray/removeFromArray.js
```diff
@@ -1,9 +1,7 @@
 // we have 2 solutions here, an easier one and a more advanced one.
 // The easiest way to get an array of all of the arguments that are passed to a function
```
thatblindgeye, line 2: Actually just a quick fix to make, let's update this to:

```suggestion
// The easiest way to get an array of the rest of the arguments that are passed to a function
```

thatblindgeye (changes requested):

thatblindgeye (approved):

PR #267, opened 2022-08-06
Fibonacci solution

abdelhakim54:
propose an alternative solution to the fibonacci exercise using recursion


xandora:
We don't accept submissions to the repository unless they are improvements to the exercises or updates to the solution branch. If you wish to push your work to GitHub you should push to your own fork, there is no need to create Pull Requests as well.

PR #276, opened 2022-09-05
Added an iterative approach with (for of)

alejandroh9800:

-   [x ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [ 11_getTheTitles: Added an iterative approach with ( for of)] The title of this PR follows the `location of change: brief description of change` format, e.g. `01_helloWorld: Update test cases`

<hr>

**1. Because:**

Closes #XXXXX

**2. This PR:

- Adds another solution to the problem.

**

### DIFF
--- a/getTheTitles/getTheTitles.js
+++ b/getTheTitles/getTheTitles.js
@@ -3,3 +3,14 @@ const getTheTitles = function(array) {
 };
 
 module.exports = getTheTitles;
+
+
+const getTheTitles = function (array) {
+  let titles = [];
+
+  for (const element of array) titles.push(element.title);
+
+  return titles;
+};
+
+module.exports = getTheTitles


alejandroh9800 (commented):
Previously I had used forEach(), but for of is a faster and more dynamic approach.

thatblindgeye (commented):
If we were to include an alternate approach, we would want to comment out the alternative method and not add another export.

That said, considering the surrounding context of where this exercise is to be done by users, I don't think this alternative approach is entirely needed. This exercise is mentioned in a lesson that goes over the map method, so it would be more beneficial for users to really be shown that method as the solution rather than a for of loop. Will leave this open for now, though, in case you have any other thoughts on it.

thatblindgeye:
Based on my above comment and no further discussion, will be closing this.

PR #319, opened 2023-01-17
09 Palindromes: Non Regex Solution Added

Hifilo:
## Because

Since the readme didn't suggest regex, I got the impression we were to do this wihtout regex. But after finishing, and checking the solution, there was only a regex version.

## This PR

- Regex Method commented
-Non Regex method with comments attached.

## Issue

Closes #XXXXX

## Pull Request Requirements

-   [x ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [x] If this PR includes changes that needs to be updated on the `solutions` branch, I have created another PR (and linked it to this PR).

### DIFF
--- a/palindromes/palindromes.js
+++ b/palindromes/palindromes.js
@@ -23,7 +23,7 @@ const palindromes = function (string) {
 	return cleanedString === reversedString;
 };
 
-//check with console.log statements below in your daw/
+//check with console.log statements below in your editor/
 
 // console.log(palindromes('A car, a man, a maraca.'));
 // console.log(palindromes('ZZZZ car, a man, a maracazzzz.'));
--- a/palindromes/palindromes.js
+++ b/palindromes/palindromes.js
@@ -1,11 +1,33 @@
-const palindromes = function(string) {
-  const processedString = string.toLowerCase().replace(/[^a-z]/g, "");
-  return (
-    processedString
-      .split("")
-      .reverse()
-      .join("") == processedString
-  );
+//This is a method using Regular Expressions, and is probably the easiest if you know how to use them
+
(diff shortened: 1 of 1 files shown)


rlmoser99 (changes requested):
Thank you for taking the time to make this update. Since we do not teach regex in Foundations or truly recommend it in our curriculum, I believe your solution should replace the current solution.

I know it has been quite some time since you submitted this PR, but would you be willing to make some changes? In addition to removing the original solution, could you remove the comments. I believe your solution is very readable to understand.

Hifilo:
> Thank you for taking the time to make this update. Since we do not teach regex in Foundations or truly recommend it in our curriculum, I believe your solution should replace the current solution.
> 
> I know it has been quite some time since you submitted this PR, but would you be willing to make some changes? In addition to removing the original solution, could you remove the comments. I believe your solution is very readable to understand.

Hi, no worries. Yes it has been a minute, and I forgot about this. 

To be fair, I wanted to use a regex method. but because it wasn't instructed to do so, I thought I was supposed to find the solution without it. 
So when I checked the solutions after I passed all tests, I was hoping to see a cleaner method, than what I had originally written. I saw the regex method which kind of left me a little perplexed. Thats really the only reason I added the solution.
Another user from another discord suggested putting the alphabet in a variable, and using filter, and that led to this solution. 

I've never gone through this review process before, so I'm not sure if I did it correctly, but I updated the code with the changes you requested, and sent another commit.

rlmoser99 (approved):
This looks great! Thanks for taking the time to come back to this PR and make these change!

PR #356, opened 2023-05-11
Fix issue with returning incorrect value for string inputs

### LINKED ISSUE #192: Fibonacci solution bug
Current fibonacci solution returns `1` instead of `0` when passed `"0"`, because it uses strict equality.

https://github.com/TheOdinProject/javascript-exercises/blob/8150be944bc01cb3bea5b50a760ae88e405693c1/fibonacci/fibonacci.js#L3

I'm opening this issue because tests state that the function should accept a string representation of a number.

Nitinbhojsingh:
Issue Fix #192 

I fixed the issue in the fibonacci function where it was returning 1 instead of 0 for string inputs that represent the number 0. I did this by adding a line of code that converts the input to a number and then checking if it's NaN or less than 0. This change should make the function work correctly for both number inputs and string inputs that represent valid numbers.

### DIFF
--- a/10_fibonacci/solution/fibonacci-solution.js
+++ b/10_fibonacci/solution/fibonacci-solution.js
@@ -1,5 +1,6 @@
 const fibonacci = function (count) {
-  if (count < 0) return "OOPS";
+  count = Number(count);
+  if (isNaN(count) || count < 0) return "OOPS";
   if (count === 0) return 0;
   let a = 0;
   let b = 1;


Nitinbhojsingh (commented):
I have recently made some changes to the fibonacci function in my forked repository, and I would like to request a review from you.

10_fibonacci/solution/fibonacci-solution.js
```diff
@@ -1,5 +1,6 @@
 const fibonacci = function (count) {
-  if (count < 0) return "OOPS";
+  count = Number(count);
```
thatblindgeye, line 2: Nit: we could assign this to a new variable instead of mutating the `count` arg

thatblindgeye (commented):
Looks like there's a merge conflict that needs resolving

thatblindgeye:
Closing due to inactivity.

PR #359, opened 2023-05-22
Removed array syntax in multiply test

fruddenfeldt:
The test for the 'multiply' function used array brackets [] for the input parameters, which caused the test to return an error when rest parameters (...args) are used in the function, like so:

const multiply = function(...args){
  return args.reduce((acc, cur) => acc * cur);
}

## Issue

Closes #XXXXX

## Pull Request Requirements

-   [ ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [ ] The title of this PR follows the `location of change: brief description of change` format, e.g. `01_helloWorld: Update test cases`
-   [ ] The `Because` section summarizes the reason for this PR
-   [ ] The `This PR` section has a bullet point list describing the changes in this PR
-   [ ] If this PR includes changes that needs to be updated on the `solutions` branch, I have created another PR (and linked it to this PR).

### DIFF
--- a/08_calculator/calculator.spec.js
+++ b/08_calculator/calculator.spec.js
@@ -40,11 +40,11 @@ describe('sum', () => {
 
 describe('multiply', () => {
 	test.skip('multiplies two numbers', () => {
-		expect(calculator.multiply([2,4])).toBe(8);
+		expect(calculator.multiply(2,4)).toBe(8);
 	});
 
 	test.skip('multiplies several numbers', () => {
-		expect(calculator.multiply([2,4,6,8,10,12,14])).toBe(645120);
+		expect(calculator.multiply(2,4,6,8,10,12,14)).toBe(645120);
 	});
 });
 


codyloyd:
This is a good idea  I think.

Can you also do a PR to the solutions branch?  the solution we have is geared toward these tests, and expects an array as a parameter.. but none of the other base-math functions (add, subtract, etc.) expect an array, so it makes sense to be consistent here.

fruddenfeldt:
Thanks. Done - see this PR: https://github.com/TheOdinProject/javascript-exercises/pull/364

cats256:
> This is a good idea I think.
> 
> Can you also do a PR to the solutions branch? the solution we have is geared toward these tests, and expects an array as a parameter.. but none of the other base-math functions (add, subtract, etc.) expect an array, so it makes sense to be consistent here.

If that's the case, we should also modify the test for the sum function in calculator.spec.js to expect rest parameters instead of expecting an array parameter. Same thing with the calculator-solution.spec.js file, in which the test for sum and multiply functions expects an array instead of rest parameters.

cats256:
I opened a PR

PR #368, opened 2023-07-04
Update sumAll-solution.js swap algorithm to standard method

cats256:
## Because
I believe this way of swapping the values between variables in JavaScript is better for learners to know about since it is more common and reinforces understanding on array destructuring

## This PR
Change the swapping algorithm to the standard way of swapping in JavaScript using array restructuring. Edit: Change for loop condition as well to the standard way

## Pull Request Requirements
-   [x] If this PR includes changes that needs to be updated on the `solutions` branch, I have created another PR (and linked it to this PR).

### DIFF
--- a/05_sumAll/solution/sumAll-solution.js
+++ b/05_sumAll/solution/sumAll-solution.js
@@ -1,7 +1,7 @@
 const sumAll = function (min, max) {
   if (!Number.isInteger(min) || !Number.isInteger(max)) return "ERROR";
   if (min < 0 || max < 0) return "ERROR";
-    if (min > max) {
+  if (min > max) {
     const temp = min;
     min = max;
     max = temp;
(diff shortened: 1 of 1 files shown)


05_sumAll/solution/sumAll-solution.js
```diff
@@ -1,13 +1,10 @@
 const sumAll = function (min, max) {
   if (!Number.isInteger(min) || !Number.isInteger(max)) return "ERROR";
   if (min < 0 || max < 0) return "ERROR";
-  if (min > max) {
-    const temp = min;
-    min = max;
-    max = temp;
-  }
+  if (min > max) [min, max] = [max, min];
+
```
thatblindgeye, line 5: Not entirely sure if we should use this as the TOP solution, just because I don't believe it's something introduced to users at the point they're asked to do this exercise (believe it's the lesson they're first introduced to arrays).

Maybe we could add a comment that shows this method as an alternative.

cats256: Done

thatblindgeye (approved):

PR #369, opened 2023-07-04
Reformat test parameters for consistency

cats256:
## Because
Reformat test parameters allow for more consistency

## This PR
Modified the test for the sum function in calculator.spec.js, as well as the sum and multiply functions in calculator-solution.spec.js file to expect rest parameters instead of expecting an array parameter. Modified the sum function in calculator-solution.js to work with the new test for the sum function. There was no need to modify the multiply function in calculator-solution.js as it already expects rest parameters

### DIFF
--- a/08_calculator/calculator.spec.js
+++ b/08_calculator/calculator.spec.js
@@ -2,55 +2,55 @@ const calculator = require('./calculator');
 
 describe('add', () => {
 	test('adds 0 and 0', () => {
-		expect(calculator.add(0,0)).toBe(0);
+		expect(calculator.add(0, 0)).toBe(0);
 	});
 
 	test.skip('adds 2 and 2', () => {
-		expect(calculator.add(2,2)).toBe(4);
+		expect(calculator.add(2, 2)).toBe(4);
 	});
 
 	test.skip('adds positive numbers', () => {
-		expect(calculator.add(2,6)).toBe(8);
+		expect(calculator.add(2, 6)).toBe(8);
 	});
 });
 
 describe('subtract', () => {
 	test.skip('subtracts numbers', () => {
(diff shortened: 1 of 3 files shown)


cats256:
@CouchofTomato hello I don't believe the maintainer is active in reviewing PR anymore right now

thatblindgeye:
@cats256 We definitely should align how the Odin solutions are written (either with rest params or passing in arrays). My only issue with using rest param instead of passing in an array for the exercise is I don't believe we really go over rest param other than as an additional resource (which isn't required reading).

I feel like if we want to go forward with this update, it may be worth moving "Fundamentals Part 5" additional resource to the assignment, so users will at least understand what the rest parameter is (this would be done in our curriculum repo).

The alternative is revert the update made to the multiply solution to accept an array.

cats256:
> @cats256 We definitely should align how the Odin solutions are written (either with rest params or passing in arrays). My only issue with using rest param instead of passing in an array for the exercise is I don't believe we really go over rest param other than as an additional resource (which isn't required reading).
> 
> I feel like if we want to go forward with this update, it may be worth moving "Fundamentals Part 5" additional resource to the assignment, so users will at least understand what the rest parameter is (this would be done in our curriculum repo).
> 
> The alternative is revert the update made to the multiply solution to accept an array.

Alright I made the multiply solution to accept an array as well as the calculator.spec.js to expect an array parameter as well (pretty much the same as calculator-solution.spec.js but with test.skip and require ./calculator instead of ./calculator-solution). calculator-solution.spec.js is no longer changed since the original tests expect an array instead of rest parameters for the multiply solution.

thatblindgeye (approved):

PR #370, opened 2023-07-04
Update fibonacci-solution.js

cats256:
More optimized solution with O(1) memory usage

### DIFF
--- a/10_fibonacci/solution/fibonacci-solution.js
+++ b/10_fibonacci/solution/fibonacci-solution.js
@@ -4,13 +4,11 @@ const fibonacci = function(count) {
 
     let firstPrev = 1;
     let secondPrev = 0;
-
-    // For clarification: curr stands for current. This is standard syntax
     
     for (let i = 2; i <= count; i++) {
-        let curr = firstPrev + secondPrev;
+        let current = firstPrev + secondPrev;
         secondPrev = firstPrev;
-        firstPrev = curr;
+        firstPrev = current;
     }
 
     return firstPrev;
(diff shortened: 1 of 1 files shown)


10_fibonacci/solution/fibonacci-solution.js
```diff
@@ -1,10 +1,17 @@
...
+    let first_prev = 1;
+    let second_prev = 0;
+
+    for (let i = 2; i <= count; i++) {
+        let curr = first_prev + second_prev;
+        second_prev = first_prev;
+        first_prev = curr;
+    }
+
+    return first_prev;
```
01zulfi, line 14: Can we go with `camelCase` please to align with what we teach?

01zulfi (changes requested):
I like the proposed solution. IMO it's easier to understand. Just have one small nit below

10_fibonacci/solution/fibonacci-solution.js
cats256: Done

10_fibonacci/solution/fibonacci-solution.js
```diff
@@ -1,10 +1,19 @@
...
+
+    let firstPrev = 1;
+    let secondPrev = 0;
+
+    // For clarification: curr stands for current. This is standard syntax
+    
+    for (let i = 2; i <= count; i++) {
+        let curr = firstPrev + secondPrev;
+        secondPrev = firstPrev;
+        firstPrev = curr;
```
01zulfi, line 13: Just one more small change please. Can we go for `current` instead of `curr` and remove the comment? This exercise is fairly early in the curriculum and I want the variable names to be as clear as possible for the learners. Thanks!

01zulfi (changes requested):

10_fibonacci/solution/fibonacci-solution.js
cats256: Done

01zulfi (approved):
LGTM :rocket:

PR #423, opened 2024-01-15
Exercise/03 reverse string: Add alternative solution as comment

### LINKED ISSUE #410: 03_reverseString: Add alternative solution
-   [x] The title of this issue follows the `location for request: brief description of request` format, e.g. `Exercises: Add exercise on XYZ`

The following checkbox is OPTIONAL:

-   [x] I would like to be assigned this issue to work on it

<hr>

**1. Description of the Feature Request:**

Hello there!
I think it will be useful for TOP learners to see an alternative solution for this exercise that is, in my opinion, quite important.
The proposal is to add another solution, perhaps in the same file as _reverseString-solution.js_  or in a new file.

The solution I am proposing is this.
```
const reverseString = function(input) {
    let result = "";
    for (let i = input.length - 1; i >= 0; i--) {
        result += input[i];
    }
    return result;
};
```

The benefit is that the learners can get to learn a style of algorithm, that is closer to what they will eventually may need to learn from data structures and algorithms (if they have not yet!). As there are many ways to solve the problem, I think it's also good that students can get exposure to different solutions and learn about the difference in their runtime as well.

jwsoh07:
## Because
It provides an alternative programming style of using basic syntax to complete the exercise (Exercise/03 reverse string) as opposed to string methods and it results in a benefit of having a faster runtime of about 3x faster.

## This PR

- Added a commented alternative solution

## Issue
Closes #410

## Pull Request Requirements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vscode
 node_modules/
 .DS_Store
+.tool-versions
--- a/03_reverseString/reverseString.js
+++ b/03_reverseString/reverseString.js
@@ -1,5 +1,15 @@
-const reverseString = function() {
+const reverseString = function(input) {
+    // approach 1.
+    return input.split('').reverse().join('');
+};
 
+const reverseStringFaster = function(input) {
+    // approach 2.
+    let result = "";
+    for (let i = input.length - 1; i >= 0; i--) {
+        result += input[i];
+    }
+    return result;
 };
 
 // Do not edit below this line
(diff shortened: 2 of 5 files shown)


thatblindgeye (changes requested):
There are some files that have updates that need to be reverted. We only want to update the solution js file for this.

jwsoh07:
Noted. I will create another PR for this one.

PR #439, opened 2024-03-03
09_palindromes: Use const for solution variable

MaoShizhong:
## Because
One of the variables uses `let` as a carryover from the original solution adapted from.
This variable is never reassigned, so would be better declared with `const`.
A tiny pedantic change but since the rest of the function's non-reassigned variables are declared with `const`, this keeps things consistent and prevents any "why use let here and const there?" confusions from learners.

## This PR
- Changes a non-reassigned variable to be declared via `const` instead of `let` 

## Issue
N/A

## Pull Request Requirements


JoshDevHub (approved):
Seems reasonable to me.

PR #440, opened 2024-03-11
Update fibonacci-solution.js: add recursive solution

AdishtJaglan:
Added the recursive approach to the solution.

## Because

This PR also add a recursive solution to the question at hand. It might create some curiousity in the mind of the reader, which might lead to greater interest. 

## This PR

- Added comments which include a recursive solution, complementing the iterative solution.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/10_fibonacci/solution/fibonacci-solution.js
+++ b/10_fibonacci/solution/fibonacci-solution.js
@@ -31,4 +31,16 @@ const fibonacci = function(countArg) {
 // }
 // return fib[count];
 
+// Another way to do this is by using recursion, define the base condition as n=0,n=1
+// if (n < 0) {
+//         return "OOPS";
+//     }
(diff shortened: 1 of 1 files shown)


MaoShizhong:
@AdishtJaglan Thank you for submitting this PR.

Introducing recursion here would not be pedagogically appropriate.
This is likely to cause more confusion and questions for learners than beneficial knowledge.
It also risks opening a huge rabbit hole into recursion way before it's appropriate to dive into that topic.

Recursion is taught later on in either full-stack pathway at a time that is more appropriate within the course. At this point, learners can then benefit from the structured guidance the course provides.

Diving into recursion without guidance and at a much earlier point in the course where learners will have far less experience with the tools and concepts in general, that doesn't seem like the best idea to me.

PR #441, opened 2024-03-22
New exercise | Factorial

nikitarevenco:
## Because

## This PR

- Adds exercise 13

## Issue

## Pull Request Requirements

### DIFF
--- a/13_factorial/README.md
+++ b/13_factorial/README.md
@@ -1,0 +1,13 @@
+# Exercise 13 - Factorial
+
+Create a function that takes an input integer greater than or equal to 1 and returns the product of all its predecessing integers. If 0 is input it returns 1. It should accept strings that when converted to a number are valid inputs. For invalid inputs it returns `undefined`.
+
+For example:
+
+```javascript
+factorial(2); // 2 * 1, Output: 2
+factorial('4'); // 4 * 3 * 2 * 1, Output: 24
+factorial(5); // 5 * 4 * 3 * 2 * 1, Output: 120
+factorial(7.2); // Output: undefined
+factorial(0); // Output: 1
+```
--- a/13_factorial/factorial.spec.js
+++ b/13_factorial/factorial.spec.js
@@ -13,7 +13,7 @@ describe('factorial', () => {
   test.skip('15th factorial number is 1307674368000', () => {
     expect(factorial(15)).toBe(1307674368000);
   });
-  test.skip('25th factorial number is 1.5511210043330986e+255', () => {
+  test.skip('25th factorial number is 1.5511210043330986e+25', () => {
     expect(factorial(25)).toBe(1.5511210043330986e+25);
   });
   test.skip('0th factorial number is 1', () => {
--- a/13_factorial/factorial.spec.js
+++ b/13_factorial/factorial.spec.js
@@ -1,0 +1,40 @@
+const factorial = require("./factorial");
+
+describe('factorial', () => {
+  test('4th factorial number is 24', () => {
+    expect(factorial(4)).toBe(24);
+  });
+  test.skip('6th factorial number is 720', () => {
+    expect(factorial(6)).toBe(720);
+  });
+  test.skip('10th factorial number is 3628800', () => {
+    expect(factorial(10)).toBe(3628800);
+  });
+  test.skip('15th factorial number is 1.3076744e+12', () => {
+    expect(factorial(15)).toBe(1.3076744e12);
+  });
+  test.skip('25th factorial number is 1.551121e+25', () => {
+    expect(factorial(25)).toBe(1.551121e25);
+  });
+  test.skip('0th factorial number is 1', () => {
+    expect(factorial(0)).toBe(1);
+  });
+  test.skip('doesn\'t accept negatives', () => {
+    expect(factorial(-25)).toBe(undefined);
+  });
+  test.skip('doesn\'t accept floats', () => {
+    expect(factorial(5.4)).toBe(undefined);
+  });
+  test.skip('DOES accept strings', () => {
+    expect(factorial("0")).toBe(1);
+  });
+  test.skip('DOES accept strings', () => {
+    expect(factorial("1")).toBe(1);
+  });
+  test.skip('DOES accept strings', () => {
+    expect(factorial("2")).toBe(2);
+  });
+  test.skip('DOES accept strings', () => {
+    expect(factorial("8")).toBe(40320);
+  });
+});
--- a/13_factorial/factorial.spec.js
+++ b/13_factorial/factorial.spec.js
@@ -1,4 +1,4 @@
-const factorial = require("./factorial");
+const factorial = require("./factorial-solution");
 
 describe('factorial', () => {
   test('4th factorial number is 24', () => {
@@ -10,11 +10,11 @@ describe('factorial', () => {
   test.skip('10th factorial number is 3628800', () => {
     expect(factorial(10)).toBe(3628800);
   });
-  test.skip('15th factorial number is 1.3076744e+12', () => {
-    expect(factorial(15)).toBe(1.3076744e12);
+  test.skip('15th factorial number is 1307674368000', () => {
+    expect(factorial(15)).toBe(1307674368000);
   });
-  test.skip('25th factorial number is 1.551121e+25', () => {
-    expect(factorial(25)).toBe(1.551121e25);
+  test.skip('25th factorial number is 1.5511210043330986e+255', () => {
+    expect(factorial(25)).toBe(1.5511210043330986e+25);
   });
   test.skip('0th factorial number is 1', () => {
     expect(factorial(0)).toBe(1);
--- a/13_factorial/solution/factorial-solution.spec.js
+++ b/13_factorial/solution/factorial-solution.spec.js
@@ -4,37 +4,37 @@ describe('factorial', () => {
   test('4th factorial number is 24', () => {
     expect(factorial(4)).toBe(24);
   });
-  test.skip('6th factorial number is 720', () => {
+  test('6th factorial number is 720', () => {
     expect(factorial(6)).toBe(720);
   });
-  test.skip('10th factorial number is 3628800', () => {
+  test('10th factorial number is 3628800', () => {
     expect(factorial(10)).toBe(3628800);
   });
-  test.skip('15th factorial number is 1.3076744e+12', () => {
-    expect(factorial(15)).toBe(1.3076744e12);
+  test('15th factorial number is 1307674368000', () => {
+    expect(factorial(15)).toBe(1307674368000);
   });
-  test.skip('25th factorial number is 1.551121e+25', () => {
-    expect(factorial(25)).toBe(1.551121e25);
+  test('25th factorial number is 1.5511210043330986e+255', () => {
+    expect(factorial(25)).toBe(1.5511210043330986e+25);
   });
-  test.skip('0th factorial number is 1', () => {
+  test('0th factorial number is 1', () => {
     expect(factorial(0)).toBe(1);
   });
-  test.skip('doesn\'t accept negatives', () => {
+  test('doesn\'t accept negatives', () => {
     expect(factorial(-25)).toBe(undefined);
   });
-  test.skip('doesn\'t accept floats', () => {
+  test('doesn\'t accept floats', () => {
     expect(factorial(5.4)).toBe(undefined);
   });
-  test.skip('DOES accept strings', () => {
+  test('DOES accept strings', () => {
     expect(factorial("0")).toBe(1);
   });
-  test.skip('DOES accept strings', () => {
+  test('DOES accept strings', () => {
     expect(factorial("1")).toBe(1);
   });
-  test.skip('DOES accept strings', () => {
+  test('DOES accept strings', () => {
     expect(factorial("2")).toBe(2);
   });
-  test.skip('DOES accept strings', () => {
+  test('DOES accept strings', () => {
     expect(factorial("8")).toBe(40320);
   });
 });
--- a/13_factorial/solution/factorial-solution.spec.js
+++ b/13_factorial/solution/factorial-solution.spec.js
@@ -4,37 +4,37 @@ describe('factorial', () => {
   test('4th factorial number is 24', () => {
     expect(factorial(4)).toBe(24);
   });
-  test('6th factorial number is 720', () => {
+  test.skip('6th factorial number is 720', () => {
     expect(factorial(6)).toBe(720);
   });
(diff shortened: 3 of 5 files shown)


MaoShizhong (changes requested):
**EDIT**: The more I think about it, the more I feel that accepting strings is unnecessary for the intent of this exercise. The point of the exercise is to practise recursion, and it's the first exercise for that to boot. Therefore, learners should be focusing on wrestling with the recursion part, not type checking and various additional unrelated things to prevent edge cases like `factorial([4])` being valid without an `Array.isArray` rejection. Accepting only numbers allows for only `!Number.isInteger(n)` and `n < 0` rejections. Far more relevant and intuitive, and does not reduce complexity for the actual point of the task, which is recursion.

---

Additional scenarios that need to be tested:
- Accept integers with `.0` e.g. `5.0` or `'5.0'`. String would be better as we'd only need a single test case for that (needing to convert it to a number would mean it also effectively tests `5.0`)
- Reject string floats e.g. `'8.9'`
- Reject some other invalid value e.g. `'foo'`

Though having said that, I'm wondering if there's much actual value in accepting string values at all. Since the exercise is primarily about recursion, I'd probably argue that it makes more sense to only accept numbers (so `'4'` or `'4.0'` would be rejected). That would also solve edge cases like `factorial([])` and `factorial([7])` being the equivalents of `factorial(0)` and `factorial(7)` respectively, as arrays coerce to strings which then cleanly convert to legit numbers in those two cases.

nikitarevenco:
> **EDIT**: The more I think about it, the more I feel that accepting strings is unnecessary for the intent of this exercise. The point of the exercise is to practise recursion, and it's the first exercise for that to boot.
> 
> Therefore, learners should be focusing on wrestling with the recursion part, not type checking and various additional unrelated things to prevent edge cases like `factorial([4])` being valid without an `Array.isArray` rejection. Accepting only numbers allows for only `!Number.isInteger(n)` and `n < 0` rejections. Far more relevant and intuitive, and does not reduce complexity for the actual point of the task, which is recursion.
> 
> Additional scenarios that need to be tested:
> 
> * Accept integers with `.0` e.g. `5.0` or `'5.0'`. String would be better as we'd only need a single test case for that (needing to convert it to a number would mean it also effectively tests `5.0`)
> * Reject string floats e.g. `'8.9'`
> * Reject some other invalid value e.g. `'foo'`
> 
> Though having said that, I'm wondering if there's much actual value in accepting string values at all. Since the exercise is primarily about recursion, I'd probably argue that it makes more sense to only accept numbers (so `'4'` or `'4.0'` would be rejected). That would also solve edge cases like `factorial([])` and `factorial([7])` being the equivalents of `factorial(0)` and `factorial(7)` respectively, as arrays coerce to strings which then cleanly convert to legit numbers in those two cases.

I agree that it would be better to make the learner not have to deal with strings and instead focus on the recursion aspect. Should this approach be employed for all future exercises?

nikitarevenco:
> **EDIT**: The more I think about it, the more I feel that accepting strings is unnecessary for the intent of this exercise. The point of the exercise is to practise recursion, and it's the first exercise for that to boot.
> 
> Therefore, learners should be focusing on wrestling with the recursion part, not type checking and various additional unrelated things to prevent edge cases like `factorial([4])` being valid without an `Array.isArray` rejection. Accepting only numbers allows for only `!Number.isInteger(n)` and `n < 0` rejections. Far more relevant and intuitive, and does not reduce complexity for the actual point of the task, which is recursion.
> 
> Additional scenarios that need to be tested:
> 
> * Accept integers with `.0` e.g. `5.0` or `'5.0'`. String would be better as we'd only need a single test case for that (needing to convert it to a number would mean it also effectively tests `5.0`)
> * Reject string floats e.g. `'8.9'`
> * Reject some other invalid value e.g. `'foo'`
> 
> Though having said that, I'm wondering if there's much actual value in accepting string values at all. Since the exercise is primarily about recursion, I'd probably argue that it makes more sense to only accept numbers (so `'4'` or `'4.0'` would be rejected). That would also solve edge cases like `factorial([])` and `factorial([7])` being the equivalents of `factorial(0)` and `factorial(7)` respectively, as arrays coerce to strings which then cleanly convert to legit numbers in those two cases.

Also, instead of making it return `undefined` lets have the learner make it throw an error! The curriculum does not officially cover how to `throw`, but they can look it up. They should know this is something that can be done.

MaoShizhong:
@nikitarevenco I think that an exercise should focus on challenging a learner on things related to its intent.

For example, this exercise's main intents are:
- Factorial
- Recursive

The recursion part is obvious. The factorial part requires we only handle valid values (non-negative integers). Therefore, the exercise should be made to challenge on those parts.

Strings can be converted to numbers before being passed in as arguments - it does not need to be handled by the function itself. Adding that capability in is an extra that's unrelated to the exercise's purpose, as it doesn't challenge the learner on any of the above key concepts (factorial/recursion). Those are what we want to challenge the learner on with this exercise, not type conversion.

MaoShizhong:
For future reference as well, if you want to increase the complexity of an exercise, you do so by making the relevant problem itself more challenging.

If you try to do so by adding in arbitrary unrelated extras, that doesn't make the exercise more challenging in a beneficial way given what its purpose is. That's just arbitrarily putting in extra stuff for no reason other than "to make it more complex".

nikitarevenco:
> For future reference as well, if you want to increase the complexity of an exercise, you do so by making the relevant problem itself more challenging.
> 
> If you try to do so by adding in arbitrary unrelated extras, that doesn't make the exercise more challenging in a beneficial way given what its purpose is. That's just arbitrarily putting in extra stuff for no reason other than "to make it more complex".

Cool! I updated it.

nikitarevenco:
> For future reference as well, if you want to increase the complexity of an exercise, you do so by making the relevant problem itself more challenging.
> 
> If you try to do so by adding in arbitrary unrelated extras, that doesn't make the exercise more challenging in a beneficial way given what its purpose is. That's just arbitrarily putting in extra stuff for no reason other than "to make it more complex".

Shall I start working on these now -

- contains() from the JS challenges
- totalIntegers() from the JS challenges
- sumSquares() from the JS challenges
- #flatten from the Ruby challenges

13_factorial/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Write a recursive function that takes a non-negative integer, and returns the product of all positive integers less than or equal to the input integer. An input of `0` should return `1`. The function should only accept numbers, so `'4'` should not be accepted as it is a string. All invalid inputs should return `undefined`.
+
+For example:
+
+```javascript
+factorial('4'); // Output: undefined
+factorial(5); // 5 * 4 * 3 * 2 * 1, Output: 120
+factorial(7.2); // Output: undefined
+factorial(0); // Output: 1
```
MaoShizhong, line 11:
```suggestion
factorial(5); // 5 * 4 * 3 * 2 * 1, Output: 120
factorial(0); // Output: 1
factorial(7.2); // Output: undefined
factorial('4'); // Output: undefined
```
Nit: I think this order makes more sense for the example.

13_factorial/factorial.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+  test.skip('doesn\'t accept negatives', () => {
+    expect(factorial(-25)).toBe(undefined);
+  });
+  test.skip('doesn\'t accept floats', () => {
+    expect(factorial(5.4)).toBe(undefined);
+  });
+  test.skip('doesn\'t accept strings', () => {
+    expect(factorial('5')).toBe(undefined);
+  });
```
MaoShizhong, line 30: As opposed to naming the test "doesn't accept strings", perhaps "only accepts numbers"?
And perhaps add a second test with the same description that passes in like, an array or some other non-number value?

13_factorial/solution/factorial-solution.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+  test.skip('doesn\'t accept negatives', () => {
+    expect(factorial(-25)).toBe(undefined);
+  });
+  test.skip('doesn\'t accept floats', () => {
+    expect(factorial(5.4)).toBe(undefined);
+  });
+  test.skip('doesn\'t accept strings', () => {
+    expect(factorial('5')).toBe(undefined);
+  });
```
MaoShizhong, line 30: The same would need to be applied here too.

MaoShizhong (changes requested):

MaoShizhong:
> > For future reference as well, if you want to increase the complexity of an exercise, you do so by making the relevant problem itself more challenging.
> > If you try to do so by adding in arbitrary unrelated extras, that doesn't make the exercise more challenging in a beneficial way given what its purpose is. That's just arbitrarily putting in extra stuff for no reason other than "to make it more complex".
> 
> Shall I start working on these now -
> 
> * contains() from the JS challenges
> * totalIntegers() from the JS challenges
> * sumSquares() from the JS challenges
> * #flatten from the Ruby challenges

I would say start working on the next intended JS exercise, and include a link to this PR in that one's text.

I think it would be most sensible to only add these exercises to the repo once we've agreed on the state of all proposed exercises, then we can merge them en masse. I'll probably discuss with the team what the best plan of action would be, but for the time being, this feels more sensible than merging an exercise that won't be used by the curriculum until all exercises are merged anyway.

13_factorial/factorial.spec.js
nikitarevenco: > As opposed to naming the test "doesn't accept strings", perhaps "only accepts numbers"? And perhaps add a second test with the same description that passes in like, an array or some other non-number value? Or a 2nd `expect` within that test block.

How about

```js
  describe('only accepts numbers', () => {
    test('doesn\'t accept negatives', () => {
      expect(factorial(-25)).toBe(undefined);
    });
    test('doesn\'t accept floats', () => {
      expect(factorial(5.4)).toBe(undefined);
    });
    test('doesn\'t accept strings', () => {
      expect(factorial('5')).toBe(undefined);
    });
    test('doesn\'t accept arrays', () => {
      expect(factorial([5])).toBe(undefined);
    });
    test('doesn\'t accept objects', () => {
      expect(factorial({})).toBe(undefined);
    });
    test('doesn\'t accept NaN', () => {
      expect(factorial(NaN)).toBe(undefined);
    });
  });
```

MaoShizhong: I think that's a bit much - perhaps just something like rejecting `'5'`, and something like `'foo'` or `[5]` is more than sufficient for testing rejection of non-numbers.
Negatives and floats can be their own tests.

(Also `-25`, `5.4` and `NaN` are numbers, so wouldn't fit that describe block anyway).

13_factorial/README.md
```diff
@@ -0,0 +1,12 @@
+# Exercise 13 - Factorial
+
+Write a recursive function that takes a non-negative integer, and returns the product of all positive integers less than or equal to the input integer. An input of `0` should return `1`. The function should only accept numbers, so `'4'` should not be accepted as it is a string. All invalid inputs should return `undefined`.
```
JoshDevHub, line 3: How do we feel about having a link in here? Maybe to something like https://simple.wikipedia.org/wiki/Factorial

13_factorial/factorial.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  test.skip('25th factorial number is 1.5511210043330986e+25', () => {
+    expect(factorial(25)).toBe(1.5511210043330986e+25);
+  });
+  test.skip('0th factorial number is 1', () => {
+    expect(factorial(0)).toBe(1);
+  });
+  test.skip('doesn\'t accept negatives', () => {
+    expect(factorial(-25)).toBe(undefined);
+  });
+  test.skip('doesn\'t accept floats', () => {
```
JoshDevHub, line 25: I'll cede to whatever JS people think is common practice, but for my tastes, double quoting has much better readability compared to escaping.

JoshDevHub (commented):
More general comments rather than things I think for sure should be included. Curious to know what both of you think @nikitarevenco @MaoShizhong 

I also agree with Mao on not having the functions necessarily be flexible with strings and stuff. I realize the other exercises in this repo often do things like that, but this is just to give people practice with typecasting data. It's not how functions tend to be written in the "real world", and it's also not particularly relevant for the task at hand with this group of exercises (wrapping your mind around recursion).

13_factorial/README.md
MaoShizhong: Agree with this

13_factorial/factorial.spec.js
MaoShizhong: And also agree with this

nikitarevenco:
@MaoShizhong @JoshDevHub 

I think once all of these new exercises get merged it wouldn't make sense to keep them separate from the project so I propose to also convert the tasks in the project into TDD exercises too and remove the project entirely

13_factorial/README.md
```diff
@@ -0,0 +1,12 @@
+# Exercise 13 - Factorial
+
+Write a [recursive factorial function](https://simple.wikipedia.org/wiki/Factorial) that takes a non-negative integer, and returns the product of all positive integers less than or equal to the input integer. An input of `0` should return `1`. The function should only accept numbers, so `'4'` should not be accepted as it is a string. All invalid inputs should return `undefined`.
```
MaoShizhong, line 3:
```suggestion
Write a recursive [factorial](https://simple.wikipedia.org/wiki/Factorial) function that takes a non-negative integer, and returns the product of all positive integers less than or equal to the input integer. An input of `0` should return `1`. The function should only accept numbers, so `'4'` should not be accepted as it is a string. All invalid inputs should return `undefined`.
```
Nit: Though the article does mention recursion, it's not about recursion nor about programming functions, only "factorial" itself so the link text ought to reflect that exactly.

(**love** simple wiki!)

13_factorial/factorial.spec.js
```diff
@@ -0,0 +1,37 @@
+const factorial = require("./factorial-solution");
```
MaoShizhong, line 1:
```suggestion
const factorial = require("./factorial");
```

MaoShizhong (changes requested):

MaoShizhong:
Just letting you know, I haven't forgotten about these. I plan to get to reviewing these this week when I have more time at my computer.

MaoShizhong (dismissed):
:fire:

@JoshDevHub Anything additional thoughts before you get a Ruby translation for stuff going?

Will get these merged once all have been approved.

JoshDevHub:
> @JoshDevHub Anything additional thoughts before you get a Ruby translation for stuff going?
> 
> Will get these merged once all have been approved.

I think it's looking pretty good. Should have some time this week to work on Ruby translations.

PR #442, opened 2024-03-23
New exercise | Contains

nikitarevenco:
## Because

## This PR

- Adds exercise 14

## Issue

## Pull Request Requirements

### DIFF
--- a/14_contains/README.md
+++ b/14_contains/README.md
@@ -1,0 +1,3 @@
+# Exercise 13 - contains
+
+Description of the exercise goes here.
--- a/14_contains/contains.spec.js
+++ b/14_contains/contains.spec.js
@@ -1,0 +1,15 @@
+const contains = require('./contains');
+
+describe('contains', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(contains()).toBe('');
+  });
+  
+  test.skip('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(contains()).toBe('');
+  });
+});
--- a/14_contains/solution/contains-solution.js
+++ b/14_contains/solution/contains-solution.js
@@ -1,0 +1,6 @@
+const contains = function() {
+  // Replace this comment with the solution code
+};
+  
+// Do not edit below this line
+module.exports = contains;
--- a/14_contains/solution/contains-solution.spec.js
+++ b/14_contains/solution/contains-solution.spec.js
@@ -1,0 +1,15 @@
+const contains = require('./contains-solution');
+
+describe('contains', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(contains()).toBe('');
+  });
+  
+  test('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(contains()).toBe('');
+  });
+});
--- a/14_contains/contains.js
+++ b/14_contains/contains.js
@@ -1,0 +1,6 @@
+const contains = function() {
+  
+};
+  
+// Do not edit below this line
+module.exports = contains;


14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  test('does not do type conversions', () => {
+    expect(contains(nestedObject, '44')).toBe(false);
+  });
+
+  test('false if property is not in object', () => {
+    expect(contains(nestedObject, 'bar')).toBe(false);
+  });
+
+  test('true if property is duplicated', () => {
+    expect(contains(nestedObject, NaN)).toBe(true);
```
MaoShizhong, line 39: `NaN` is better tested in its own test because NaN equality is a special case (`NaN === NaN` evaluates to `false` by definition) and would require specific `NaN` equality checks, or a method that has that build in (like `.includes`) e.g.
```js
test('true if NaN is a value within the object', () => {
  expect(contains(object, NaN)).toBe(true);
});
```

A `true if provided value is within the object, even if duplicated` test (wording following previous point about test descriptions) would better use a normal primitive, meaning you should change the test object to duplicate something different.

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,45 @@
+const contains = require('./contains-solution');
+
+describe('contains', () => {
+  const nestedObject = {
```
MaoShizhong, line 4:
```suggestion
  const object = {
```
`nestedObject` implies *that* object is nested inside something, but it's not.

MaoShizhong (changes requested):
I imagine you're leaving out `contains.spec.js` to make it easier to do changes to only one file instead of two, but just a reminder in case that wasn't your intention and you just forgot!

MaoShizhong:
@nikitarevenco to make things easier to know when to do things, once you've made any requested changes etc., if you could re-request a review from whichever maintainer, that would make it much easier to know when it's okay to review (so we don't go to review and submit comments right before you push another commit that changes things again)

14_contains/solution/contains-solution.js
```diff
@@ -0,0 +1,12 @@
+const contains = function(current, search) {
+  if (Object.values(current).includes(search)) return true;
+
+  for (const nestedObject of Object.values(current).filter((value) => typeof value === 'object')) {
+    return contains(nestedObject, search)
+  }
+  
+  return false
+};
```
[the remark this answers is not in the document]
nikitarevenco: You learn something new everyday! Your changes definitely improve the readability. 

I think it's quite nice that the explicit `NaN` check is not required, since `includes` handles that for us gracefully. So how about keeping the original idea and adding a comment that explains this detail and explains that usually an explicit `NaN` check would be required? 

Also, yeah I'm going to paste the changes into the `.spec` files once it looks like the tests won't change anymore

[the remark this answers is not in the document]
MaoShizhong: I think that sounds reasonable

nikitarevenco:
> @nikitarevenco to make things easier to know when to do things, once you've made any requested changes etc., if you could re-request a review from whichever maintainer, that would make it much easier to know when it's okay to review (so we don't go to review and submit comments right before you push another commit that changes things again)

Alright, that makes sense!

nikitarevenco:
> @nikitarevenco to make things easier to know when to do things, once you've made any requested changes etc., if you could re-request a review from whichever maintainer, that would make it much easier to know when it's okay to review (so we don't go to review and submit comments right before you push another commit that changes things again)

I've opened the next exercise for review! I wanted to 'request' - but seems like that cannot be done on a PR with no reviews yet. For the future, do you prefer me to add a comment when the next PR is ready for review or?

MaoShizhong:
@nikitarevenco No need to comment or request a review when un-drafting something. The un-drafting itself is enough to signify that something is ready for review.

I'm not reviewing these because I'm assigned them, nor am I the only one allowed to review them. I just happen to be seeing these and choosing to add my reviews. Any of the maintainers may wish to review (or not) at any time. And I'd probably be wanting at least another opinion before they're all finalised for merging.

14_contains/solution/contains-solution.js
```diff
@@ -0,0 +1,16 @@
+const contains = function(obj, searchValue) {
+  const values = Object.values(obj);
+
+  // NaN !== NaN, so we would usually have to do an explicit check with Math.isNaN to test for it. However, Array.prototype.includes handles that for us
```
MaoShizhong, line 4:
```suggestion
  // NaN === NaN evaluates to false (as per the IEEE 754 standard)
  // Normally, we would have to do an explicit Number.isNaN() check to compare NaN equality
  // However, Array.prototype.includes automatically handles this for us
```
Nit: Personally, I think something like this is a little clearer as to what the problem is (NaN === NaN being false), what the usual resolution would be, and then how our solution works. What are your thoughts?

More importantly, it's `Number.isNaN` as opposed to `Math.isNaN`

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+  test('true if provided string is within the object, even if duplicated', () => {
+    expect(contains(object, 'e')).toBe(true);
+  });
+
+  test('false if the provided value exists but is not a primitive', () => {
+    expect(contains(object, [42])).toBe(false);
+  });
+
+  test('true if NaN is a property within the object', () => {
```
MaoShizhong, line 47:
```suggestion
  test('true if NaN is a value within the object', () => {
```

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,50 @@
...
+                banana: NaN,
+                moreStuff: {
+                    something: 'foo',
+                    answer: [42]
+                }
+              }
+          }
+      }
+    }
+  }
```
JoshDevHub, line 21: I think a good edge case to add here would be a `null` value somewhere in the object. The solution you have doesn't handle this, so it'd also require some tinkering with that.

MaoShizhong: Great catch

JoshDevHub: `typeof null === 'object'` javascript is a beautiful language :smirk:

14_contains/solution/contains-solution.js
nikitarevenco: I think this is a good change, however I don't think it would be necessary to include the IEEE bit. If we *do* then it may be worth linking to the wiki page

MaoShizhong: You could leave a simple link, though it won't be like markdown because it's a JS comment, so will just be a bare link.
The IEEE part isn't particularly necessary, more so the structure of the comments IMO

14_contains/solution/contains-solution.spec.js
nikitarevenco: Yeah you always have to be careful checking `typeof === null`, this is what happens when you make a language in just 10 days xD

14_contains/solution/contains-solution.js
JoshDevHub, line 4: I'm maybe a bit wary of this, but I'll defer to the JS people on it. Generally I'd say code that demands a commented explanation is a bit of a warning sign, but `NaN` behavior is one of those things that's so weird, maybe it'd need a comment no matter how it's written.

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,55 @@
+const contains = require('./contains-solution');
+
+describe('contains', () => {
+  const object = {
```
JoshDevHub, line 4: I think it could also be nice to add a case that makes sure the algorithm can properly handle sibling branches and isn't all just children. Can maybe just reorganize the object here to cover this.

You can see an example of what I mean by looking at the original solution on the [site currently linked in the curriculum](https://www.codingame.com/playgrounds/5422/js-interview-prep-recursion). Their solution is:
```js
function contains(obj, value) {
  for(var key in obj){
  if(typeof obj[key] === 'object'){
      return contains(obj[key], value);
    }

    if (obj[key] === value){
      return true;
    }
  }
  return false;
}
```
If you pass a nested object that's just all children, it's fine
```js
const object = {
  data: {
    childOne: {
      foo: "bar",
      childTwo: {
        number: 42
      }
    }
  }
}

// true
console.log(contains(object, 42));
```

But if you pass it an object where there are "sibling" branches (ie nested objects on the same level), you'll see there's a logical flaw.

```js
const objectWithSiblings = {
  data: {
    childOne: {
      foo: "bar",
    },
    childTwo: {
      number: 42
    }
  }
}

// outputs `false` but should be `true`!
console.log(contains(objectWithSiblings, 42)
```

Your solution naturally handles this just fine, but I think the tests should look for it because it's a common problem people can run into that would sneak through currently.

JoshDevHub (commented):
A few more comments.

14_contains/solution/contains-solution.js
```diff
@@ -0,0 +1,16 @@
...
+  if (values.includes(searchValue)) return true;
+
+  const nestedObjects = values.filter((value) => typeof value === 'object' && value !== null);
+  for (const nestedObject of nestedObjects) {
+    return contains(nestedObject, searchValue);
+  }
+
+  return false;
```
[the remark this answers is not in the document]
MaoShizhong: Not even just a nit - I just realised that specifically due to the test object not testing for sibling objects (since it's all just children), it all passes despite us returning on the first iteration of the for..of loop.

The for..of approach would need to be
```js
for (const nestedObject of nestedObjects) {
  if (contains(nestedObject, searchValue)) {
    return true;
  }
}
```
which the `Array.prototype.some` approach solves just fine when accounting for values present in non-first nested objects.

I think `.some` is appropriate for where people would be in this part of the course in JS :+1:

14_contains/solution/contains-solution.js
MaoShizhong: The `NaN === NaN` returning `false` thing will definitely throw a lot of people off, I can imagine. It's not something people would've come across much by this point but will be a good time to learn about that.

That's why I personally would write a specific NaN check in, though I'm also not opposed to using `.includes`. The comment there is mainly as a teaching point that `.includes` has built-in `NaN` checking in case anyone had to do an explicit NaN check but then saw the provided solution didn't.
I imagine Ruby's `nil` behaves a little more...intuitively?

JoshDevHub: Oh right you are. I was thinking the way it was currently written could handle sibling branches. Nice spot.

JoshDevHub: I commented on the wrong thread ... but you get my point

nikitarevenco: Thanks!

nikitarevenco: yeah js can be a little counter-intuitive at times

14_contains/solution/contains-solution.spec.js
nikitarevenco: Nicely spotted! My solution did not actually work with this new requirement, so I had to spend some time thinking how to fix it. Its been a while since I used recursion or queues so this was certainly a nice refresh

14_contains/solution/contains-solution.js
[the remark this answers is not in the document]
nikitarevenco: I have slightly re-written the solution to use a queue now so this suggestion is no longer applicable to the new approach (with a queue)

14_contains/solution/contains-solution.js
```diff
@@ -1,19 +1,22 @@
...
     (value) => typeof value === "object" && value !== null,
   );
-  for (const nestedObject of nestedObjects) {
-    return contains(nestedObject, searchValue);
-  }
 
-  return false;
+  const newQueue = queue.concat(nestedObjects);
+
+  return contains(null, searchValue, newQueue);
```
JoshDevHub, line 19: I think `some()` is a lot simpler than using a queue and should pass all the specs. Also, "queue" is not something learners are aware about while doing these exercises. It's not introduced until the Common Data Structures and Algorithms lesson, which comes after these exercises.

```js
const contains = function(object, searchValue) {
  const values = Object.values(object);
  
  if (values.includes(searchValue)) return true;
  
  const nestedObjects = values.filter((value) => typeof value === "object" && value !== null);
  return nestedObjects.some((nestedObject) => contains(nestedObject, searchValue));
}
```

14_contains/solution/contains-solution.js
```diff
@@ -1,19 +1,22 @@
...
+  const item = queue.shift();
+  const values = Object.values(item);
 
   // NaN === NaN evaluates to false
   // Normally, we would have to do an explicit Number.isNaN() check to compare NaN equality
   // However, Array.prototype.includes automatically handles this for us
   if (values.includes(searchValue)) return true;
 
   const nestedObjects = values.filter(
+    // typeof null === 'object' evaluates to true ¯\_(ツ)_/¯
```
JoshDevHub, line 13: :laughing:

14_contains/solution/contains-solution.js
nikitarevenco: Oh yeah, thanks - you solution is certainly a lot cleaner. I have never considered the `some` method but turns out that it is actually really useful.

14_contains/README.md
```diff
@@ -0,0 +1,3 @@
+# Exercise 14 - contains
+
+Write a function that searches for a value in a nested object. It returns true if the object contains that value.
```
MaoShizhong, line 3: 1. I think we should make it explicit that objects are compared by reference.
2. Like other exercises, some basic examples will help.

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    expect(contains(object, "bar")).toBe(false);
+  });
+
+  test("true if provided string is within the object, even if duplicated", () => {
+    expect(contains(object, "e")).toBe(true);
+  });
+
+  test("false if the provided value exists but is not a primitive", () => {
+    expect(contains(object, [42])).toBe(false);
+  });
```
MaoShizhong, line 46: I think we shouldn't make this exclusion for "not a primitive".
This test passes (with `false`) because `[42] === [42]` is false. If you passed in a reference to the same object as one in the nested object, it'd return `true` (no modifications to the current solution).

So we should modify the tests so instead of `[42]`, we define it separately and have `answer: meaningOfLifeArray` or something. Then, we have a test for `contains(object, meaningOfLifeArray)` and `contains(object, [42])` respectively.

14_contains/contains.spec.js
MaoShizhong, line 1: Once the other things are addressed, if you can duplicate the tests to this file, that'd be great.

MaoShizhong (changes requested):
@JoshDevHub Any further thoughts yourself?

14_contains/README.md
nikitarevenco: Ok, I added some basic examples. If someone wants to view all cases they can always look at the tests

14_contains/solution/contains-solution.spec.js
nikitarevenco: Ok, I added this

14_contains/contains.spec.js
nikitarevenco: I'll do that at once will all the pull requests once they're all marked as ready to go :+1:

JoshDevHub (commented):
Looks good I think :+1:

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    expect(contains(object, "e")).toBe(true);
+  });
+
+  test("false if the provided array has the same structure as an array within the object, however the reference is not the same", () => {
+    expect(contains(object, [42])).toBe(false);
+  });
+
+  test("true if the provided array is a reference to an array that exists within the object", () => {
+    expect(contains(object, meaningOfLifeArray)).toBe(true);
+  });
```
MaoShizhong, line 51: Some nits:

- I'd put the same reference test first
- Perhaps a little less wordy on the test descriptions? e.g.
  - "true if the object contains the same object by reference"
  - "false if no matching object reference"

14_contains/contains.spec.js
MaoShizhong, line 1: Once the remaining test changes are implemented, you can copy them over and skip all but the first. Once that's all done, I'm happy with this one :fire:

14_contains/solution/contains-solution.spec.js
nikitarevenco: sounds good!!

14_contains/contains.spec.js
nikitarevenco: Ok, I copied them over!

14_contains/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Objects are compared by reference.
+
+Examples:
+
+```javascript
+contains({ foo: "foo" }, "bar") // true
```
MaoShizhong, line 10:
```suggestion
contains({ foo: "foo" }, "bar") // false
```

MaoShizhong (changes requested):

MaoShizhong (approved):

14_contains/solution/contains-solution.spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+  test("false if no matching object reference", () => {
+    expect(contains(object, [42])).toBe(false);
+  });
+
+  test("true if NaN is a value within the object", () => {
+    expect(contains(object, NaN)).toBe(true);
+  });
+
+  test("false if the provided value exists and is null", () => {
```
JoshDevHub, line 57:
```suggestion
  test("true if the provided value exists and is null", () => {
```
Shouldn't this be `true` here?

JoshDevHub (changes requested):
Sorry for the random bump @nik-rev but I've started working on the Ruby exercises so we can finally get this merged. I noticed one small nit while working on the Ruby path PR for this exercise.

PR #443, opened 2024-03-23
New exercise | Total Integers

nikitarevenco:
## Because

## This PR

- Adds exercise 15

## Issue

## Pull Request Requirements

### DIFF
--- a/15_totalIntegers/solution/totalIntegers-solution.js
+++ b/15_totalIntegers/solution/totalIntegers-solution.js
@@ -1,0 +1,6 @@
+const totalIntegers = function() {
+  // Replace this comment with the solution code
+};
+  
+// Do not edit below this line
+module.exports = totalIntegers;
--- a/15_totalIntegers/solution/totalIntegers-solution.spec.js
+++ b/15_totalIntegers/solution/totalIntegers-solution.spec.js
@@ -1,0 +1,15 @@
+const totalIntegers = require('./totalIntegers-solution');
+
+describe('totalIntegers', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(totalIntegers()).toBe('');
+  });
+  
+  test('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(totalIntegers()).toBe('');
+  });
+});
--- a/15_totalIntegers/README.md
+++ b/15_totalIntegers/README.md
@@ -1,0 +1,3 @@
+# Exercise 13 - totalIntegers
+
+Description of the exercise goes here.
--- a/15_totalIntegers/totalIntegers.js
+++ b/15_totalIntegers/totalIntegers.js
@@ -1,0 +1,6 @@
+const totalIntegers = function() {
+  
+};
+  
+// Do not edit below this line
+module.exports = totalIntegers;
--- a/15_totalIntegers/totalIntegers.spec.js
+++ b/15_totalIntegers/totalIntegers.spec.js
@@ -1,0 +1,15 @@
+const totalIntegers = require('./totalIntegers');
+
+describe('totalIntegers', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(totalIntegers()).toBe('');
+  });
+  
+  test.skip('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(totalIntegers()).toBe('');
+  });
+});


MaoShizhong (changes requested):
General but important comment to consider for this one - I'll let you have a think about how you want to approach this.

1. No need to reinvent the wheel for [isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger)
1. You only test nested arrays but what if there are objects within the array(s) that contain integer values?
e.g.
```js
[4, 6, { a: 1, b: { a: 10, b: 11 } }, 9]
```

nikitarevenco:
> General but important comment to consider for this one - I'll let you have a think about how you want to approach this.
> 
> 1. No need to reinvent the wheel for [isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger)
> 2. You only test nested arrays but what if there are objects within the array(s) that contain integer values?
>    e.g.
> 
> ```js
> [4, 6, { a: 1, b: { a: 10, b: 11 } }, 9]
> ```

Didn't know that method exists, how convenient!

Good point about the objects with numbers, that certainly makes the challenge more interesting

MaoShizhong:
> Didn't know that method exists, how convenient!

[Pssst...](https://github.com/TheOdinProject/javascript-exercises/blob/8b2ed42f9a8547dbea98420d5892dbac02ef0231/13_factorial/solution/factorial-solution.js#L2)

nikitarevenco:
> > Didn't know that method exists, how convenient!
> 
> [Pssst...](https://github.com/TheOdinProject/javascript-exercises/blob/8b2ed42f9a8547dbea98420d5892dbac02ef0231/13_factorial/solution/factorial-solution.js#L2)

Wait whaaaat xD

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,15 @@
+const totalIntegers = function(obj, count = 0) {
+
+  if (typeof obj !== 'object' || obj === null) return;
+
+  const elements = Array.isArray(obj) ? obj : Object.values(obj);
```
MaoShizhong, line 5:
```suggestion
  const elements = Object.values(obj);
```
Object.values() coerces to an object, so passing it an array will just have `Object.values(obj)` return the array itself.

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,15 @@
...
+
+  for (const el of elements) {
+    if (Number.isInteger(el)) ++count;
+    if (typeof el === 'object' && el !== null) count += totalIntegers(el);
```
MaoShizhong, line 9:
```suggestion
    if (Number.isInteger(el)) {
      count++;
    } else if (typeof el === 'object' && el !== null) {
      count += totalIntegers(el);
    }
```
I believe post-fix increment is consistent with the curriculum's lessons and exercises.

`else if` would be more semantic here, even if `if` achieves the same result.
It also means the conditions can be chained.
Especially when chaining, having curly braces IMO is far more readable as the indentation makes it clearer to see what the code to run for each condition actually is.

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,15 @@
+const totalIntegers = function(obj, count = 0) {
+
+  if (typeof obj !== 'object' || obj === null) return;
```
MaoShizhong, line 3:
```suggestion
  if (typeof obj !== 'object' || obj === null) {
    return;
  }
```
For consistency with the braced conditions in the below loop.

MaoShizhong (commented):
Haven't had time to assess the tests but at least in terms of how the solution is written, here are some suggestions.

15_totalIntegers/solution/totalIntegers-solution.js
nikitarevenco: Ohh right, that's neat!

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,20 @@
+const totalIntegers = function(obj, count = 0) {
+
+  if (typeof obj !== 'object' || obj === null) {
```
MaoShizhong, line 3: What's your opinion on extracting this out to an `isObject` helper function, since there isn't a built-in one in JS?
e.g.
```javascript
const isObject = function(value) {
  return typeof value === 'object' && value !== null;
}
```
Then we can do 
`if (!isObject(obj))` and later, `else if (isObject(el))`.

That can also help drive it home that object checking has to involve a null check.

15_totalIntegers/solution/totalIntegers-solution.spec.js
```diff
@@ -0,0 +1,31 @@
+const totalIntegers = require('./totalIntegers-solution');
+
+describe('totalIntegers', () => {
```
MaoShizhong, line 3: I think it would be valuable to start with just a few very basic tests e.g.

```javascript
test('Works with a simple array of numbers', () => {
  expect(totalIntegers([1, 2, 3])).toBe(3);
});
test('Ignores non-integer values', () => {
  expect(totalIntegers([1, 2, '3', 4])).toBe(3);
});
test('Works with simple objects', () => {
  expect(totalIntegers({ a: 1, b: '2', c: 3 })).toBe(2);
});
```

Then afterwards we can get into the "more complex" test cases.
This helps people get the basic logic down, before then trying to work with any of the recursive sides of things.

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,20 @@
+const totalIntegers = function(obj, count = 0) {
+
```
MaoShizhong, line 2:
```suggestion
const totalIntegers = function(obj, count = 0) {
```

MaoShizhong (commented):
2 more general comments:
- Don't forget to add `.skip` to all but the 1st test in the suite.
- I know the earlier exercises use a traditional function expression but I think it may be sensible to have these new exercises use arrow functions, and also for the non-solution template. The older exercises can always be updated another time. @JoshDevHub, thoughts on having the exercises use arrow functions instead of traditional func expressions?

15_totalIntegers/solution/totalIntegers-solution.js
nikitarevenco: Sure, lets dry up the code a bit

nikitarevenco:
> 2 more general comments:
> 
> * Don't forget to add `.skip` to all but the 1st test in the suite.
> * I know the earlier exercises use a traditional function expression but I think it may be sensible to have these new exercises use arrow functions, and also for the non-solution template. The older exercises can always be updated another time. @JoshDevHub, thoughts on having the exercises use arrow functions instead of traditional func expressions?

For the first point, I am pretty sure that the `.solution.spec` files don't have any `.skip`ped tests in any of the previous exercises. But yeah that will have to be done when I end up adding all the tests to the regular spec files.

Also I agree that it would be nice to have them all be arrow functions however I think consistency is important and it would be better to update them in a different PR. Also considering that the exercise generator tool would need to be updated too it would certainly have to all be in a separate PR.

MaoShizhong:
Ah yes, the solution test suite doesn't. As long as the skips get added when you copy over for the non-solution test suite, then all is good :+1: 

Also fair point on the exercise generator tool - let's stick to trad. function expressions then.

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,21 @@
+const totalIntegers = function (obj, count = 0) {
```
JoshDevHub, line 1: With `count` not actually being passed to any recursive calls, I think it's better to just define it in the function body with a `let count = 0` somewhere.

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,21 @@
...
+  const isObject = (value) => typeof value === 'object' && value !== null;
```
JoshDevHub, line 2: This helper definition should be moved outside of the recursive function's body. This is because every time this function is run, there must be a new `isObject()` function object created for the current scope. This is wasteful -- each recursive call could just be using the same `isObject()` helper function if it were defined outside this function body.

This isn't super meaningful in the context of this specific problem (ie none of the examples are going to have performance issues because of this), but it *is* good practice to avoid defining constant factors (like this helper function) inside of loops or recursive functions.

JoshDevHub (changes requested):
Left some thoughts. Let me know if you have any questions.

MaoShizhong:
@nikitarevenco Any issues with the above comments from Josh?

nikitarevenco:
> @nikitarevenco Any issues with the above comments from Josh?

Ok, went over all the comments just now and added the suggestions

15_totalIntegers/solution/totalIntegers-solution.js
```diff
@@ -0,0 +1,24 @@
+// We could define this helper function inside of totalIntegers however that would not be best practice as everytime that the recursive totalIntegers is run it would re-create the isObject function when it could just re-use it.
```
MaoShizhong, line 1: Not sure we need this comment. I would think a quick reminder that `typeof null` is `'object'` would be sensible (it *always* catches me out).

MaoShizhong (commented):
After the comment below, I'm happy to approve once the tests and appropriate skips have been copied over to the other spec file :fire:

15_totalIntegers/solution/totalIntegers-solution.js
JoshDevHub: Yeah, I agree that this comment should maybe just focus on `typeof null` behavior

JoshDevHub (commented):
Other than Mao's one live comment, I think everything looks good with this.

PR #444, opened 2024-03-23
New exercise | sumSquares

nikitarevenco:
## Because

## This PR

- Adds exercise 16

## Issue

## Pull Request Requirements

### DIFF
--- a/16_sumSquares/README.md
+++ b/16_sumSquares/README.md
@@ -1,0 +1,3 @@
+# Exercise 13 - sumSquares
+
+Description of the exercise goes here.
--- a/16_sumSquares/solution/sumSquares-solution.js
+++ b/16_sumSquares/solution/sumSquares-solution.js
@@ -1,0 +1,6 @@
+const sumSquares = function() {
+  // Replace this comment with the solution code
+};
+  
+// Do not edit below this line
+module.exports = sumSquares;
--- a/16_sumSquares/solution/sumSquares-solution.spec.js
+++ b/16_sumSquares/solution/sumSquares-solution.spec.js
@@ -1,0 +1,15 @@
+const sumSquares = require('./sumSquares-solution');
+
+describe('sumSquares', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(sumSquares()).toBe('');
+  });
+  
+  test('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(sumSquares()).toBe('');
+  });
+});
--- a/16_sumSquares/sumSquares.js
+++ b/16_sumSquares/sumSquares.js
@@ -1,0 +1,6 @@
+const sumSquares = function() {
+  
+};
+  
+// Do not edit below this line
+module.exports = sumSquares;
--- a/16_sumSquares/sumSquares.spec.js
+++ b/16_sumSquares/sumSquares.spec.js
@@ -1,0 +1,15 @@
+const sumSquares = require('./sumSquares');
+
+describe('sumSquares', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(sumSquares()).toBe('');
+  });
+  
+  test.skip('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(sumSquares()).toBe('');
+  });
+});


nikitarevenco:
I don't think we need sumSquares exercise, it is too similar to totalIntegers

16_permutations/solution/permutations-solution.spec.js
```diff
@@ -0,0 +1,82 @@
...
+        [3, 1, 2],
+        [3, 2, 1],
+      ],
+    );
+  });
+  test('Works for array of size four', () => {
+    [actual, expected] = outcome(
+      [1, 2, 3, 4],
+      [
+        [
```
MaoShizhong, line 53: You've got an extra nested array here that's not needed, but also your `outcome` function results in the test passing despite this being an incorrect expected outcome.

That being said, even after fixing this, you don't need an `outcome` helper function with an `afterEach` that uses `.toBe`. You can just use the [`.toEqual`](https://jestjs.io/docs/expect#toequalvalue) matcher that's designed exactly for this purpose.

```javascript
test('Works for array of size two', () => {
  expect(permutations([1, 2])).toEqual([
    [1, 2],
    [2, 1],
  ]);
});
```
Having all the tests as per the above format will all pass (and will also fail if you keep the unintended extra nested array in the last test "expected" array.

MaoShizhong (changes requested):
Haven't looked at the the solution yet but one (technically 2) important change needed for the test suite below.

16_permutations/solution/permutations-solution.spec.js
nikitarevenco: > You've got an extra nested array here in the "expected array" that's not needed, but also your `outcome` function results in the test passing despite this being an incorrect expected outcome.
> 
> That being said, even after fixing this, you don't need an `outcome` helper function with an `afterEach` that uses `.toBe`. You can just use the [`.toEqual`](https://jestjs.io/docs/expect#toequalvalue) matcher that's designed exactly for this purpose.
> 
> ```js
> test('Works for array of size two', () => {
>   expect(permutations([1, 2])).toEqual([
>     [1, 2],
>     [2, 1],
>   ]);
> });
> ```
> 
> Having all the tests as per the above format will all pass (and will also fail if you keep the unintended extra nested array in the last test "expected" array.

Yes but the problem with that is, what happens if the learner's solution returns the correct array but in a different order?

> You've got an extra nested array here in the "expected array" that's not needed, but also your `outcome` function results in the test passing despite this being an incorrect expected outcome.
> 
> That being said, even after fixing this, you don't need an `outcome` helper function with an `afterEach` that uses `.toBe`. You can just use the [`.toEqual`](https://jestjs.io/docs/expect#toequalvalue) matcher that's designed exactly for this purpose.
> 
> ```js
> test('Works for array of size two', () => {
>   expect(permutations([1, 2])).toEqual([
>     [1, 2],
>     [2, 1],
>   ]);
> });
> ```
> 
> Having all the tests as per the above format will all pass (and will also fail if you keep the unintended extra nested array in the last test "expected" array.

OHhhhhhhh wow, I had no idea the following test would pass:
```javascript
    expect([1, 2]).toEqual([2, 1])
```

MaoShizhong: Fair point I missed about the `toEqual` order. That can be remedied by sorting both sides
```javascript
test('Works for array of size two', () => {
  expect(permutations([1, 2]).sort()).toEqual(
    [
      [1, 2],
      [2, 1],
    ].sort()
  );
});
```
Note the `.sort()` for the expect is after the permutations call, not on the array passed into the permutations call.
Do that, and the order of the matched array won't matter.

16_permutations/solution/permutations-solution.js
```diff
@@ -0,0 +1,19 @@
...
+
+  const newPerms = [];
+  perms.forEach((el) => {
+    const missing = original.filter((item) => !el.includes(item));
+    missing.forEach((itemMissing) => newPerms.push([...el, itemMissing]));
+  });
+
+  if (newPerms.every((el) => el.length === original.length)) return newPerms;
+  return permutations(original, newPerms);
+};
```
MaoShizhong, line 16: I'll be honest, I had quite a bit of trouble understanding this solution as a whole. Not because the method itself is bad, I actually think it's really elegant. My difficult primarily came from the naming of things, as several of the things aren't really named very accurately. Hence it threw me off as I was expecting some things to represent one thing, but what they actually represented (and thus how they were ultimately used) was something very different.

Once I clocked what they actually represented, the approach made total sense.

`currentPermutations` to me, indicates something that contains all actual permutations at this point in time. So for an input array of `[1, 2, 3, 4]`, a `currentPermutations` array containing `[[1, 2, 3, 4], [1, 2, 4, 3]]` makes sense to me. Each element is a permutation, and the whole array contains only the permutations currently found.

Instead, what you're doing with each recursive call, is building up an array of **partial permutations**. For example, `[1]` and `[1, 4]` are both partial permutations of `[1, 2, 3, 4]`, but the `currentPermutations` naming implies they're full permutations.

The below is a suggestion, but implements the above comment's changes as well as different naming. Checking `.every` for the end base case is fine, but by nature of this approach building up partial permutations one element at a time, all elements will be the same length as each other at every recursive call. Also no strong opinion on `perm`/`permutation`, just personally prefer the full. Let me know what you think.

```js
const permutations = function (original, partialPermutations = original.map((num) => [num])) {
  if (original.length === 0) return [];
  if (original.length === 1) return [original];

  const newPartialPermutations = [];
  partialPermutations.forEach((partialPermutation) => {
    const missingNums = original.filter((num) => !partialPermutation.includes(num));
    missingNums.forEach((missingNum) => {
      newPartialPermutations.push([...partialPermutation, missingNum]);
    });
  });

  if (newPartialPermutations[0].length === original.length) {
    return newPartialPermutations;
  }
  return permutations(original, newPartialPermutations);
};
```

MaoShizhong (changes requested):

16_permutations/solution/permutations-solution.js
MaoShizhong: I should also note that the solution will not work if the input array has duplicates.
You may prefer to not include this additional complexity, so if that's the case, the instructions should also make clear that the learner will not need to account for duplicate numbers.

16_permutations/solution/permutations-solution.js
```diff
@@ -0,0 +1,19 @@
+const permutations = function (original, currentPermutations) {
+  if (original.length === 1) return original;
+
+  const perms = currentPermutations
+    ? currentPermutations
+    : original.map((el) => [el]);
```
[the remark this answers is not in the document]
nikitarevenco: > ```suggestion
> const permutations = function (original, currentPermutations = original.map((num) => [num])) {
>   if (original.length === 0) return [];
>   if (original.length === 1) return [original];
> ```
> 1. Based on the intent of lines 4-6, instead of the ternary, it'd be clearer to write `const perms = currentPermutations ?? original.map((num) => [num]);`. `||` can be used as well, but the intent of that line really is just if `currentPermutations` doesn't exist, therefore `??` is more semantic. *But*, since that's the only intent of that line, you should just make it a default parameter.
> 1. `el` renamed to `num` to match the instructions. If you wish for the instructions to not be limited to numbers, then `el` can be kept and the instructions amended.
> 1. For an input of `[1]`, it wouldn't make sense to return `[1]` but rather, `[[1]]`. The function returns an array of permutations. The only permutation of `[1]` is `[1]`, so an array containing that would be `[[1]]`, as opposed to `[1]`, which implies the permuted array isn't an array but a number. This is consistent with the other examples, where an array of permuted arrays is returned. The relevant test will need to be updated accordingly.
> 1. If you're going to do a shortcut for 1-length arrays, then you should probably also include one for empty arrays. The above suggestion treats empty arrays as having no permutations, thus returning only an empty array. An alternative, would be to treat an empty array as having exactly one permutation (one way to arrange emptiness). IF that's the case, the two ifs can be combined to just `if (original.length < 2) return [original]`, so that an input of an empty array returns `[[]]`. However you decide to treat empty array inputs, you should make it clear in the instructions how they should be treated, and include a test for that.

1. I like your idea in your proposed code with the default parameters, however I feel like this could be a good case to show the ?? operator, which the learners are unlikely to use and this seems like a good use case of that.

1. I think its best to keep it to numbers not to overcomplicate this exercise, so renaming el to num is a good call

1. Ohh yeah, cant believe I missed that. Thanks!

1. `0!` evaluates to `1`, so a set with no elements will always have a single way to arrange it.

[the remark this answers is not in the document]
MaoShizhong: Regarding point 1, we really don't need a new variable to maybe be assigned the parameter value. `??` would only be used in that scenario. The equivalent without a new variable would be
```js
partialPermutations ??= original.map((num) => [num]);
```

I'm more in favour of the default param, as I think that's more semantic for this specific scenario (value for parameter if it's initialised `undefined`). I would never personally set a default param via `??=`, I'd rather use the tool meant for that purpose. So personally, I think if we wanted to demonstrate the use of `??=`, it wouldn't be here.

16_permutations/solution/permutations-solution.js
nikitarevenco: > I'll be honest, I had quite a bit of trouble understanding this solution as a whole. Not because the method itself is bad, I actually think it's really elegant (like, *very*). My difficulty primarily came from the naming of things, as several of the things aren't really named very accurately. Hence it threw me off as I was expecting some things to represent one thing, but what they actually represented (and thus how they were ultimately used) was something very different.
> 
> Once I clocked what they actually represented, the approach made total sense.
> 
> `currentPermutations` to me, indicates something that contains all actual permutations at this point in time. So for an input array of `[1, 2, 3, 4]`, a `currentPermutations` array containing `[[1, 2, 3, 4], [1, 2, 4, 3]]` makes sense to me. Each element is a permutation, and the whole array contains only the permutations currently found.
> 
> Instead, what you're doing with each recursive call, is building up an array of **partial permutations**. For example, `[1]` and `[1, 4]` are both partial permutations of `[1, 2, 3, 4]`, but the `currentPermutations` naming implies they're full permutations.
> 
> The below is a suggestion, but implements the above comment's changes as well as different naming. Checking `.every` for the end base case is fine, but by nature of this approach building up partial permutations one element at a time, all elements will be the same length as each other at every recursive call. Also no strong opinion on `perm`/`permutation`, just personally prefer the full. Let me know what you think.
> 
> ```js
> const permutations = function (original, partialPermutations = original.map((num) => [num])) {
>   if (original.length === 0) return [];
>   if (original.length === 1) return [original];
> 
>   const newPartialPermutations = [];
>   partialPermutations.forEach((partialPermutation) => {
>     const missingNums = original.filter((num) => !partialPermutation.includes(num));
>     missingNums.forEach((missingNum) => {
>       newPartialPermutations.push([...partialPermutation, missingNum]);
>     });
>   });
> 
>   if (newPartialPermutations[0].length === original.length) {
>     return newPartialPermutations;
>   }
>   return permutations(original, newPartialPermutations);
> };
> ```
> 
> 

This def makes the code a lot more readable. Misnaming a single variable can lead to so much confusion, I'll certainly take a moment to take this in as probably misnamed variables are harder to understand than a variable that has a random meaningless name lol.

I agree that `partialPermutations` makes significantly more sense than `currentPermutations`, thank you for pointing this out.

16_permutations/solution/permutations-solution.js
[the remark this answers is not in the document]
nikitarevenco: > Regarding point 1, we really don't need a new variable to maybe be assigned the parameter value. `??` would only be used in that scenario. The equivalent without a new variable would be
> ```js
> partialPermutations ??= original.map((num) => [num]);
> ```
> 
> I'm more in favour of the default param, as I think that's more semantic for this specific scenario (value for parameter if it's initialised `undefined`). I would never personally set a default param via `??=`, I'd rather use the tool meant for that purpose. So personally, I think if we wanted to demonstrate the use of `??=`, it wouldn't be here.
> 
> Also, `??` is introduced in Foundations somewhere in a javascript.info article, IIRC. I'm sure people will come across it at one point. It's a somewhat recent operator anyway, so it's not like it's replacing anything insanely ridiculous.
> 
> Regarding point 4, nice justification. In that case, we can return `[original]` when the length is `< 2`, then include a test for empty array input, and specify that expectation in the isntructions.

Also, what do you think about if we used `Set` instead of `Array` for this one? I feel like doing that makes *sense* in regards to it, since we often talk about different ways we can arrange a *set* of, consecutive integers in this example.

This would have an added benefit of introducing learners to sets, and this seems to me at least a good use case of that. Often times when we are discussing the permutations of something, we are talking about a set. A deck of cards never has duplicates for example.

Sets are cool and to me feel like a topic small enough to be introduced in an exercise's `README`

Also this would naturally exclude the option with duplicates if we decide to go this route.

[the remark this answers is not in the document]
MaoShizhong: I'll push back on using Sets instead of arrays.
Sets have their own interface to deal with and while they're no complicated by any means, learning what a Set is and how to use one isn't the point of this exercise, which is instead to recursively generate permutations of something. That should be the sole focus of the challenge.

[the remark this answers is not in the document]
nikitarevenco: Okay, sure. I understand where you are coming from and sets can be certaintly unnecessary overhead. In that case, I'm going to change this exercise so that instead of taking an array of consecutive integers, it will now take in an array of random integers, which means that there will be tests for duplication. I like your idea (in another comment, but feels like its better to respond here) and I think that certainly makes the challenge more interesting

Also, we could even expand this challenge to allow for all data types like objects, NaN, strings etc. That shouldn't be too complicated as an element that's an array would not have to have the exercise find *that* sub-array's permutations

What do you think?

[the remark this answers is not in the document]
MaoShizhong: Had a think about this and in all honesty, I think there's no need to deal with different types and duplicates. I think the complexity with just numbers (the current solution already works fine for random non-duplicated numbers so no changes are required) is just right. The main challenge will be how to generate a new permutation (or with your approach, build partial permutations num by num), and I think that should be the primary focus of the task.

MaoShizhong:
Any issues with the comments above?

16_permutations/solution/permutations-solution.spec.js
nikitarevenco: Okay, I added this

16_permutations/solution/permutations-solution.js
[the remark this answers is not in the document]
nikitarevenco: Alright so I refactored the code and added an additional test for an empty array.

16_permutations/solution/permutations-solution.js
nikitarevenco: Ok, I've made the changes
- improved variable naming
- clarified that the same integer will not be given twice in the README
- added example and tests for empty array
- refactored to use default parameter

nikitarevenco:
> Any issues with the comments above?

Okay I have added your suggestions

16_permutations/permutations.spec.js
```diff
@@ -0,0 +1,64 @@
+const permutations = require("./permutations");
+
+describe("permutations", () => {
+  test("1 possible permutation for a set containing 0 numbers", () => {
+    expect(permutations([]).sort()).toEqual([[]].sort());
+  });
+
+  test.skip("1 possible permutation for a set containing 1 number", () => {
+    expect(permutations([1]).sort()).toEqual([[1]].sort());
+  });
```
MaoShizhong, line 10: Nit and only a suggestion - I feel like omitting the `.sort()` with these makes it easier to read, and of course no change in behaviour. We only need the `sort()` for the other ones that have more than one element in the returned array.

```suggestion
  test("1 possible permutation for a set containing 0 numbers", () => {
    expect(permutations([])).toEqual([[]]);
  });

  test.skip("1 possible permutation for a set containing 1 number", () => {
    expect(permutations([1])).toEqual([[1]]);
  });
```

16_permutations/solution/permutations-solution.spec.js
```diff
@@ -0,0 +1,60 @@
+const permutations = require("./permutations-solution");
+
+describe("permutations", () => {
+  test("1 possible permutation for a set containing 0 numbers", () => {
+    expect(permutations([]).sort()).toEqual([[]].sort());
+  });
+  test("1 possible permutation for a set containing 1 number", () => {
+    expect(permutations([1]).sort()).toEqual([[1]].sort());
+  });
```
MaoShizhong, line 9: With the same `.sort()` omissions applied here

16_permutations/permutations.spec.js
nikitarevenco: yea good idea!

16_permutations/solution/permutations-solution.js
```diff
@@ -0,0 +1,28 @@
+const permutations = function (
```
JoshDevHub, line 1: Sorry for not looking at this until just now.

This is a cool solution, but I can't help but think the traditional "swap" solution for this would be better for a few different reasons.
1. The swap solution is fully built upon recursion. Yours pieces together a lot of the solution iteratively. You can kind of see what I mean by adding a `step` parameter that can effectively log the number of stack frames that are created
   ```js
   const permutations = function (
     original,
     partialPermutations = original.map((num) => [num]),
     step = 0
   )  {
     // add this line to the top of the function:
      console.log(step);|
      
      // stuff
      
      // increment on recursion:
      return permutations(original, newPartialPermutations, step + 1);
    }
    ```
    You'll see that calling `permutations([1, 2, 3, 4])` actually only creates *three* stack frames to generate all 24 permutations. This isn't necessarily a "bad" thing per se, but it does mean that this particular solution isn't very divide-and-conquer oriented.
2. The swap solution is much simpler in my opinion. This is the main reason I prefer it, and it's probably self evident from the amount of code and number of operations going on.
3. The swap solution is much faster. This isn't a huge deal either, although I suspect this exercise would get a lot of PRs from people who know the traditional "best" way to solve the problem and would rather see that as the solution.

You can see the swap solution, along with a benchmark I wrote for it against this solution with this [gist](https://gist.github.com/JoshDevHub/c0db85eb10c8c237b104c57ef13ae9d8). Let me know if all this makes sense or if you have any questions.

Also @MaoShizhong I know you're not a maintainer anymore and might not want anything to do with this work (feel free to tell me if that's the case), but I figured I'd page you anyway because I'm interested in your thoughts as well.

MaoShizhong: I don't mind being paged for second opinions, I already did my bulk of reviewing before si it's not a problem.

I like the current solution but hadn't considered your points too much, but on reconsideration, they do make sense to me and so I'd favour that i.e.
- More tied in with recursion
- Also the more traditional approach and more likely to be seen/referenced elsewhere

I can definitely see people misquestioning why this solution was chosen over the traditional, suggesting some underlying commentary that it's significantly better (which it wouldn't have been because chosen solutions are never specifically that sort of commentary, yet it's often seen as such).

nik-rev: I like this solution for the aforementioned points

PR #445, opened 2024-03-23
(17) New exercise | Flatten

nikitarevenco:
## Because

[Previous](https://github.com/TheOdinProject/javascript-exercises/pull/444)
## This PR

- Adds exercise 17

## Issue

## Pull Request Requirements

### DIFF
--- a/17_flatten/README.md
+++ b/17_flatten/README.md
@@ -1,0 +1,3 @@
+# Exercise 13 - flatten
+
+Description of the exercise goes here.
--- a/17_flatten/README.md
+++ b/17_flatten/README.md
@@ -1,6 +1,7 @@
 # Exercise 17 - flatten
 
-Write a function that takes in an array and an optional depth argument which defaults to `Infinity` if not specified.
+Write a function that takes in an array of positive integers and an optional depth argument which defaults to `Infinity` if not specified.
+
 The function returns a new array with all sub-array elements concatenated into it recursively up to the specified depth.
 
 ```javascript
--- a/17_flatten/README.md
+++ b/17_flatten/README.md
@@ -1,3 +1,10 @@
-# Exercise 13 - flatten
+# Exercise 17 - flatten
 
-Description of the exercise goes here.
+Write a function that takes in an array and an optional depth argument which defaults to `Infinity` if not specified.
+The function returns a new array with all sub-array elements concatenated into it recursively up to the specified depth.
+
+```javascript
+flatten([[1, 2], [3, 4]]); // Output: [1, 2, 3, 4]
+flatten([[1, [8, 9]], [3, 4]], 2); // Output: [1, [8, 9], 3, 4]
+flatten([2, 4, 6]); // Output: [2, 4, 6]
+```
--- a/17_flatten/flatten.js
+++ b/17_flatten/flatten.js
@@ -1,0 +1,6 @@
+const flatten = function() {
+  
+};
+  
+// Do not edit below this line
+module.exports = flatten;
--- a/17_flatten/flatten.spec.js
+++ b/17_flatten/flatten.spec.js
@@ -1,0 +1,15 @@
+const flatten = require('./flatten');
+
+describe('flatten', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(flatten()).toBe('');
+  });
+  
+  test.skip('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(flatten()).toBe('');
+  });
+});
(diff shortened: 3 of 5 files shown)


MaoShizhong (commented):
What do we think about the benefit of this exercise in context, given that [`Array.prototype.flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) exists which is identical to this, other than the default depth is 1 instead of Infinity?

While any kind of practice with recursion is still practice with recursion, I feel like if we're to create new exercises for that purpose, they ideally should be:
- problems that are naturally more suited to using some kind of recursive approach than pure iteration. An exception can definitely be made for the Factorial exercise, given it's the first one and also a classic for dipping your toe in recursion. But later exercises should probably look to fit this bill.
- something that isn't just recreating a built-in function. If we want practice with recursion, we probably would be better off solving a problem that would need us to build something recursively, as opposed to recreating a built-in where in the real world, we'd just use the built-in.

My personal opinion is that this exercise doesn't meet the second criteria, where a different exercise that does would be more appropriate while still affording the learner plenty of good practice with recursion.

@JoshDevHub tagging you for your opinion since you did mention something among these lines at some point.

nikitarevenco:
> What do we think about the benefit of this exercise in context, given that [`Array.prototype.flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) exists which is identical to this, other than the default depth is 1 instead of Infinity?
> 
> While any kind of practice with recursion is still practice with recursion, I feel like if we're to create new exercises for that purpose, they ideally should be:
> - problems that are naturally more suited to using some kind of recursive approach than pure iteration. An exception can definitely be made for the Factorial exercise, given it's the first one and also a classic for dipping your toe in recursion. But later exercises should probably look to fit this bill.
> - something that isn't just recreating a built-in function. If we want practice with recursion, we probably would be better off solving a problem that would need us to build something recursively, as opposed to recreating a built-in where in the real world, we'd just use the built-in.
> 
> My personal opinion is that this exercise doesn't meet the second criteria, where a different exercise that does would be more appropriate while still affording the learner plenty of good practice with recursion.
> 
> @JoshDevHub tagging you for your opinion since you did mention something among these lines at some point.

I honestly don't think it is a big deal that `flat` is already a built-in array method. At the end of the day, these exercises are mostly to practice recursion and this problem is fairly challenging and interesting to solve.

Take the HashMap project for example, you're not going to implement your own hashmap and probably going to use something like a `Set` in the real world. But it is a nice learning experience to kind of get a feel for how things work under the hood.

Sure, it may be better to have more exercises that focus on other tasks, and are not necessarily about re-creating a built-in, however I don't think there's any *harm* in leaving this one in. People will still learn a lot if they complete it.

Regardless, I'm content if you still think it should be removed. If we are at this point though, I think its now important to decide which exercises we *want* to add soon. 

I have added a comment to the original issue as it is certaintly more relevant there than in this PR.

https://github.com/TheOdinProject/curriculum/issues/27265#issuecomment-2041651379

MaoShizhong:
This is just my opinion though. If others disagree and think `flatten` would be perfectly fine, then by all means, let's include it. If that is to be the case, we should frame the exercise to specifically be a recreation of `Array.prototype.flat`.

The difference I find is that with the Hash Map lesson and project, you're not tasked to recreate Sets or Maps. Also, the pedagogy behind the project is the practise some lower level concepts like hashing, collisions, growth etc. To practice those, you kinda need to create a hash map from scratch (again, not a Set or Map - just a barebones hash map according to that spec). The intent of that project is very different.

In this case, the concept we want learners to practice is recursion, especially the recognition of patterns and problems that lends themselves well to recursion. Doing so is not particularly tightly bound to flattening nested arrays, so I don't see it as particularly necessary to make people recreate that built-in, when there are other exercises that can pose more unique and valuable recursive problems. Hope that helps my perspective make more sense?

nikitarevenco:
> This is just my opinion though. If others disagree and think `flatten` would be perfectly fine, then by all means, let's include it. If that is to be the case, we should frame the exercise to specifically be a recreation of `Array.prototype.flat`.
> 
> The difference I find is that with the Hash Map lesson and project, you're not tasked to recreate Sets or Maps. Also, the pedagogy behind the project is to practise some lower level concepts like hashing, collisions, growth etc. To practice those, you kinda need to create a hash map from scratch (again, not a Set or Map - just a barebones hash map according to that spec). The intent of that project is very different.
> 
> In this case, the concept we want learners to practice is recursion, especially the recognition of patterns and problems that lends themselves well to recursion. Doing so is not particularly tightly bound to flattening nested arrays, so I don't see it as particularly necessary to make people recreate that built-in, when there are other exercises that can pose more unique and valuable recursive problems. Hope that helps my perspective make more sense?
> 
> 

I see where you are coming from, thanks for clarifying it! You make a good point about the hash map project.

PR #448, opened 2024-03-24
New exercise | Tower of Hanoi

nikitarevenco:
## Because

## This PR

- Adds exercise 20

## Issue

## Pull Request Requirements

### DIFF
--- a/20_hanoi/README.md
+++ b/20_hanoi/README.md
@@ -1,0 +1,3 @@
+# Exercise 20 - hanoi
+
+Description of the exercise goes here.
--- a/20_hanoi/hanoi.spec.js
+++ b/20_hanoi/hanoi.spec.js
@@ -1,0 +1,15 @@
+const hanoi = require('./hanoi');
+
+describe('hanoi', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(hanoi()).toBe('');
+  });
+  
+  test.skip('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(hanoi()).toBe('');
+  });
+});
--- a/20_hanoi/solution/hanoi-solution.js
+++ b/20_hanoi/solution/hanoi-solution.js
@@ -1,0 +1,6 @@
+const hanoi = function() {
+  // Replace this comment with the solution code
+};
+  
+// Do not edit below this line
+module.exports = hanoi;
--- a/20_hanoi/solution/hanoi-solution.spec.js
+++ b/20_hanoi/solution/hanoi-solution.spec.js
@@ -1,0 +1,15 @@
+const hanoi = require('./hanoi-solution');
+
+describe('hanoi', () => {
+  test('First test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(hanoi()).toBe('');
+  });
+  
+  test('Second test description', () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(hanoi()).toBe('');
+  });
+});
--- a/20_hanoi/hanoi.js
+++ b/20_hanoi/hanoi.js
@@ -1,0 +1,6 @@
+const hanoi = function() {
+  
+};
+  
+// Do not edit below this line
+module.exports = hanoi;


nikitarevenco:
Ready for review!

20_hanoi/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+The rules are as follows:
+- Each disk has a length. For abstraction purposes, we will imagine each disk as an integer with that disks' length.
+- We can only move 1 disk at a time between any of the towers
+- The towers are modelled as stacks. We can move disks from the top and add them to the top with the `Array.prototype.pop` and `Array.prototype.push` methods.
+- **Disks can only be placed on top of disks that are smaller**. For instance, we cannot have the following tower: `[3, 4]`. It would violate the rules
+
+Your task is to create a function, `hanoi`, that when given an integer `n` will perform the required operations to transport all the numbers from the `first tower`
+onto the `third tower`.
```
MaoShizhong, line 24: I think this bit could be made clearer as to *exactly* what's expected from the learner from their `hanoi` function.

e.g. (wording suggestion only)
> Your task is to create a function, `hanoi`, that when given the number of disks in the starting tower, will return an array containing the state of the towers after every "move". The first element will be the towers' initial state and the last element will be the towers' final state.

**BIG** consideration (will expand in a comment on the tests file) - whether we want to enforce solving in the minimum number of moves, or if any solution that goes from start to finish shall be accepted.

20_hanoi/solution/hanoi-solution.spec.js
MaoShizhong, line 1: One issue I see with these tests is that there's no guarantee that someone's "moves" array will actually match the test, even if they successfully went from start to finish, one move at a time, and utilising recursion.

If we want to enforce that the solution should solve and return an array with the minimum number of moves (+1 of course since we have the starting position), then I *believe* the move order is indeed fixed at least to 4 disks, so what we have is fine. We'd need to make that clear in the instructions. Also note that the `.length` assertion wouldn't be needed since passing the `.toEqual` matcher implicitly checks the length also matches.

However, if we *don't* want to enforce a minimum moves solution, and only require they get from start to finish, we'd need to make that clear in the instructions and change the tests to account for different intermediary moves, as well as our conditions of disk size, e.g.:
- Assert first and last elements of moves array to be starting and ending position.
- Assert every element of the moves array is an array containing 3 arrays
- Assert no move contains a tower with disks stacked in an invalid order

Once we decide on our rules and make them explicit, I can then look at the solution and go from there.

MaoShizhong (changes requested):

20_hanoi/solution/hanoi-solution.spec.js
MaoShizhong: Speculation for sure. I don't have code in front of me that could produce such a non-minimum-moves solution that uses recursion, but remember that you've got the benefits of knowing a solution. Learners won't and who knows what they may come up with. It's entirely reasonable that they might utilise recursion somewhere in the solution that doesn't pose problems yet still allows for a different move order, including duplicate moves. As far as they'd be concerned, they utilised recursion and met the rest of the specs, yet the test fails.

Therefore, we should be explicit about the requirements. Minimum order makes sense, and I agree that *probably* many intuitively recursion solutions would result in a minimum order move list anyway. So if we want to enforce minimum order, that should be an explicit part of the instructions.

That leaves less room for what the learner considers acceptable (but might still fail tests) to differ from what we actually want.

nikitarevenco: > Speculation for sure. I don't have code in front of me that could produce such a non-minimum-moves solution that uses recursion, but remember that you've got the benefits of knowing a solution. Learners won't and who knows what they may come up with. It's entirely reasonable that they might utilise recursion somewhere in the solution that doesn't pose problems yet still allows for a different move order, including duplicate moves. As far as they'd be concerned, they utilised recursion and met the rest of the specs, yet the test fails.
> 
> Therefore, we should be explicit about the requirements. Minimum order makes sense, and I agree that _probably_ many intuitively recursion solutions would result in a minimum order move list anyway. So if we want to enforce minimum order, that should be an explicit part of the instructions.
> 
> That leaves less room for what the learner considers acceptable (but might still fail tests) to differ from what we actually want.

How about if we go with the original requirements (and make it more explicit) that the solution has to be completed in the minimum amount of steps - and if learners face issues then the requirements can be relaxed?

MaoShizhong: We can see how things go as they go, as long as we make a decision and make sure the instructions are explicit about those conditions, else your tests *technically* don't test your instructions but a stricter version of those instructions.

Original reqs sounds good, so if you can make the instructions more explicit about that, that'd be awesome. You can also then remove any of the redundant `.length` tests since you're testing for value equality with an array already

20_hanoi/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+// ...
+
+[[], [], [3, 2, 1]]
+```
+
+The rules are as follows:
+- Each disk has a length. For abstraction purposes, we will imagine each disk as an integer with that disks' length.
+- We can only move 1 disk at a time between any of the towers
+- The towers are modelled as stacks. We can move disks from the top and add them to the top with the `Array.prototype.pop` and `Array.prototype.push` methods.
```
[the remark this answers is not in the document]
nikitarevenco: Okay

20_hanoi/solution/hanoi-solution.spec.js
nikitarevenco: Let me know what you think, I've made the changes. But this is the paragraph I mostly changed:

> Your task is to create a function, `hanoi(n)`, that when given the number of disks in the starting tower (`n`), will return an array. The first element of this array will be the towers' initial state. The last element will be the towers' final state. Every intermediary element will represent a step to get from the initial position to the final position.
The function **must** return a solution in the minimum number of moves. i.e. there will be no duplicates in the array returned.

20_hanoi/README.md
```diff
@@ -0,0 +1,39 @@
+# Exercise 20 - hanoi
+
+Good job, you have come a very long way. You should be proud of yourself for making it this far. At this point, you hopefully feel comfortable
+with recursion, and so, this **final recursive exercise** should give you a proper challenge and take your abilities to the *next* level.
```
MaoShizhong, line 4:
```suggestion
Good job, you have come a very long way. You should be proud of yourself for making it this far. At this point, you hopefully feel comfortable with recursion, and so, this **final recursive exercise** should give you a proper challenge and take your abilities to the *next* level.
```

20_hanoi/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Tower of Hanoi is a classic challenge in which we have 3 towers. Tower 2 and tower 3 are empty.
+Tower 1 has `n` disks. The goal is to move all disks from tower 1 to tower 3 so that they are in the exact same order. For instance:
```
MaoShizhong, line 7:
```suggestion
Tower of Hanoi is a classic challenge in which we have 3 towers. Tower 2 and tower 3 are empty. Tower 1 has `n` disks. The goal is to move all disks from tower 1 to tower 3 so that they are in the exact same order. For instance:
```

20_hanoi/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+The rules are as follows:
+- Each disk has a length. For abstraction purposes, we will imagine each disk as an integer with that disk's length.
+- We can only move 1 disk at a time between any of the towers
+- The towers are modelled as stacks. We can move disks from the top of one stack to the top of another stack with the `Array.prototype.pop` and `Array.prototype.push` methods.
+- **Disks can only be placed on top of disks that are smaller**. For instance, we cannot have the tower `[3, 4]` but `[4, 3]` is fine.
+
+Your task is to create a function, `hanoi(n)`, that when given the number of disks in the starting tower (`n`), will return an array. The first element of this array will be the towers' initial state. The last element will be the towers' final state. Every intermediary element will represent a step to get from the initial position to the final position.
+The function **must** return a solution in the minimum number of moves. i.e. there will be no duplicates in the array returned.
```
MaoShizhong, line 24:
```suggestion
Your task is to create a function, `hanoi(n)`, that when given the number of disks in the starting tower (`n`), will return an array. The first element of this array will be the towers' initial state. The last element will be the towers' final state. Every intermediary element will represent a step to get from the initial position to the final position.

The function **must** return a solution in the minimum number of moves. i.e. there will be no duplicates in the array returned.
```

20_hanoi/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+For example, lets say we had given this function `3`, it will then output the Tower of Hanoi solution as a series of steps:
```
MaoShizhong, line 26:
```suggestion
For example, `hanoi(3)` will output the Tower of Hanoi solution as a series of steps:
```

20_hanoi/solution/hanoi-solution.js
```diff
@@ -0,0 +1,29 @@
...
+  toTowerIndex = 2,
+  towers = [
+    Array(n)
+      .fill() // Array function creates an array with "empty" slots which map does not iterate over
+      .map((_, i) => n - i),
+    [],
+    [],
+  ],
+  steps = [towers.map((tower) => [...tower])],
+) {
```
MaoShizhong, line 14: While this of course works, I feel because of stuff like this, having the function return an array of tower arrays showing the state at each step feels a little unnecessary.

I think the more common approach where you provide steps as strings (within an array) would be more intuitive, not detract from the point of the exercise being the recursive solution, and also make the provided solution *way* cleaner. I'm not particularly a fan of needing to do more complex logic in default params.

e.g.
```js
const hanoi = function (
    n,
    fromTower = 0,
    storageTower = 1,
    toTower = 2,
    steps = []
) {
    if (n <= 0) {
        return steps;
    }

    hanoi(n - 1, fromTower, toTower, storageTower, steps);

    steps.push(
        `Move disc ${n} from tower ${fromTower + 1} to tower ${toTower + 1}`
    );

    hanoi(n - 1, storageTower, fromTower, toTower, steps);

    return steps;
};
```

Then the test for `hanoi(2)` would be like
```js
  test.only("hanoi(2) should solve the puzzle in 3 moves", () => {
    const result = hanoi(2);
    expect(result).toEqual([
        'Move disc 1 from tower 1 to tower 2',
        'Move disc 2 from tower 1 to tower 3',
        'Move disc 1 from tower 2 to tower 3'
    ]);
  });
```

We can specify the template for the move string in the instructions.

MaoShizhong (changes requested):

20_hanoi/solution/hanoi-solution.js
nikitarevenco: Ok, I think this makes sense. I changed all the tests aswell.

20_hanoi/README.md
```diff
@@ -0,0 +1,38 @@
...
+[[3, 2, 1], [], []],
+[[3, 2], [], [1]],
+[[3], [2], [1]],
+[[3], [2, 1], []],
+[[], [2, 1], [3]],
+[[1], [2], [3]],
+[[1], [], [3, 2]],
+[[], [], [3, 2, 1]],
+]
+```
```
MaoShizhong, line 38: The instructions need to be amended to reflect the new requirements of returning an array of strings with the moves - the required string template must be provided as well.

20_hanoi/hanoi.spec.js
```diff
@@ -0,0 +1,65 @@
...
+
+    expect(hanoi()).toBe("");
+  });
+
+  test.skip("Second test description", () => {
+    // Replace this comment with any other necessary code, and update the expect line as necessary
+
+    expect(hanoi()).toBe("");
+  });
+});
```
MaoShizhong, line 65:
```suggestion

```

MaoShizhong (changes requested):

JoshDevHub:
I've got the other PRs mostly ready to go for the Ruby course, so hopefully we can finally get this work merged soon.

The main thing giving me pause right now is this specific exercise, which I'm not entirely sold on for a couple of reasons:
1. I think it's quite challenging, and I'm not all that interested in putting learners through the ringer on these questions. They really just need to understand recursion well enough to do DFS type things on their Binary Search Trees. I kind of want to stay away from more challenging problem types.
1. The problem setup/instructions present a more abstract situation than the other problems, which have a pretty clear and logically straightforward (input -> output) setup. With every other problem we give for these exercises, the answer can be known to the learner just from looking at the input and understanding the (quite basic) transformation we want. This one doesn't work that way. Even if you understand the rules, you can't quickly intuit what `hanoi(4)` should return. I argue that this makes the problem not great for learning recursion.

Willing to be swayed on some of this though if either of you are really set on keeping it around @mao-sz @nik-rev

PR #456, opened 2024-05-03
Exercise 4: Use const instead of var

nikitarevenco:
It's best to encourage learners to use `const` rather than `var`


MaoShizhong (approved):
Great catch

PR #457, opened 2024-05-03
Exercise 8: Remove recursive solution

nikitarevenco:
Learners would not have been taught about Recursion by this point.

A solution like this could send them down an unnecessary rabbit hole when it will be covered later on in the course.


MaoShizhong (approved):
Very strongly agreed. Didn't even realise this one was in here!
(sorry ThirtyThreeB!)

PR #458, opened 2024-05-03
Exercise 10: Add solution using `reduce` method

nikitarevenco:
I think this way of solving the problem is also pretty interesting, but is a fair bit more complicated than the previous solutions


MaoShizhong:
For similar reasons as my [comment in another PR](https://github.com/TheOdinProject/javascript-exercises/pull/463#issuecomment-2093135792), I'm going to have to push back strongly on this one.
Especially since we already have one alternative approach listed. In all honesty, adding more would just likely cause further confusion and detraction from the point of simply completing the exercises according to the spec in whatever way otherwise.

PR #459, opened 2024-05-03
Exercise 12: Make solution less verbose

nikitarevenco:
I felt like the solution is a little too verbose. I also wanted to take this moment to appreciate the `??=` operator, I think learners will find it interesting at least :>

### DIFF
--- a/12_findTheOldest/solution/findTheOldest-solution.js
+++ b/12_findTheOldest/solution/findTheOldest-solution.js
@@ -3,7 +3,7 @@ const getAge = function (birth, death) {
     death = new Date().getFullYear();
   }
   return death - birth;
-}
+};
 
 const findTheOldest = function (people) {
   return people.reduce((oldest, currentPerson) => {
--- a/12_findTheOldest/solution/findTheOldest-solution.js
+++ b/12_findTheOldest/solution/findTheOldest-solution.js
@@ -1,19 +1,17 @@
-const findTheOldest = function (array) {
-  return array.reduce((oldest, currentPerson) => {
-    const oldestAge = getAge(oldest.yearOfBirth, oldest.yearOfDeath);
-    const currentAge = getAge(
-      currentPerson.yearOfBirth,
-      currentPerson.yearOfDeath
-    );
-    return oldestAge < currentAge ? currentPerson : oldest;
-  });
+const getAge = function (person) {
+  // The nullish coalescing assignment operator
+  // evaluates to the right hand side if the left hand side is undefined
+  // Else it evaluates to the left hand side
+  person.yearOfDeath ??= new Date().getFullYear();
+
+  return person.yearOfDeath - person.yearOfBirth;
 };
 
-const getAge = function (birth, death) {
-  if (!death) {
-    death = new Date().getFullYear();
-  }
-  return death - birth;
+const findTheOldest = function (array) {
+  // Sort the people from oldest to youngest
+  return array.sort(
+    (person, nextPerson) => getAge(nextPerson) - getAge(person),
+  )[0];
 };
 
 module.exports = findTheOldest;
--- a/12_findTheOldest/solution/findTheOldest-solution.js
+++ b/12_findTheOldest/solution/findTheOldest-solution.js
@@ -17,23 +17,25 @@ const findTheOldest = function (people) {
 };
 
 // An alternative solution
-// const getAge = function (person) {
-//  // The nullish coalescing assignment operator
-//  // only does the assignment if the left side is "nullish" (evaluates to undefined or null)
-//  // if the left side has any other value, no assignment happens
-//  // here, we use ??= to set the current year for our subtraction below only if there is no year of death
-//  person.yearOfDeath ??= new Date().getFullYear();
-//
-//  return person.yearOfDeath - person.yearOfBirth;
-//};
+/* ALTERNATIVE SOLUTION
+const getAge = function (person) {
+  // The nullish coalescing assignment operator
+  // only does the assignment if the left side is "nullish" (evaluates to undefined or null)
+  // if the left side has any other value, no assignment happens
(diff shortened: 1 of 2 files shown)


12_findTheOldest/solution/findTheOldest-solution.js
```diff
@@ -1,19 +1,17 @@
...
-const getAge = function (birth, death) {
-  if (!death) {
-    death = new Date().getFullYear();
-  }
-  return death - birth;
+const findTheOldest = function (people) {
+  // Sort the people from oldest to youngest
+  return people.sort(
+    (person, nextPerson) => getAge(nextPerson) - getAge(person),
+  )[0];
```
MaoShizhong, line 14: 2 things:
1. Given this point in the curriculum, I'm not a particular fan of the `fn()[i]` pattern, and feel it would be more appropriate for learners to see things written more explicitly.
2. `people` will contain a reference to the array object passed in, and `sort` is a mutating method so while it happens to pass these tests, is not actually an appropriate solution. `toSorted` should be used here.

e.g.
```js
const findTheOldest = function (people) {
  const sortedPeopleOldestFirst = people.toSorted(
    (person, nextPerson) => getAge(nextPerson) - getAge(person)
  );
  const oldestPerson = sortedPeopleOldestFirst[0];
  return oldestPerson;
};
```
The `const oldestPerson =` line is definitely optional and you can just return `sortedPeopleOldestFirst[0]` but I like the further explicitness for learning sake.

12_findTheOldest/solution/findTheOldest-solution.js
```diff
@@ -1,19 +1,17 @@
...
-    const currentAge = getAge(
-      currentPerson.yearOfBirth,
-      currentPerson.yearOfDeath
-    );
-    return oldestAge < currentAge ? currentPerson : oldest;
-  });
+const getAge = function (person) {
+  // The nullish coalescing assignment operator
+  // evaluates to the right hand side if the left hand side is undefined
+  // Else it evaluates to the left hand side
```
MaoShizhong, line 4: I'm not sure I'd explain it this way. Unlike `??` it doesn't just "evaluate to the right hand side", as it's not just an expression but rather an assignment statement.

I'd probably say something more like
```suggestion
  // only does the assignment if the left side evaluates to undefined or null
  // if the left side has any other value, no assignment happens
  // here, we use ??= to set the current year for our subtraction below only if there is no year of death
```

MaoShizhong (changes requested):
I like this approach as a whole but personally I'd prefer it as a commented-out alternative solution, since there's absolutely nothing wrong with the current solution. A single alternative solution of relatively similar and reasonable complexity, I think is reasonable for the last exercise of the section.

Additionally, while defining `getAge` *after* `findTheOldest` works perfectly fine (because `getAge` is only actually called after it's defined) in the original, I agree it just reduces confusion to define it first. Could you also swap the order in the original solution?

12_findTheOldest/solution/findTheOldest-solution.js
nikitarevenco: Definitely agree on the explicitness, it can be harder to digest code that uses less variables!

12_findTheOldest/solution/findTheOldest-solution.js
```diff
@@ -1,5 +1,12 @@
-const findTheOldest = function (array) {
-  return array.reduce((oldest, currentPerson) => {
+const getAge = function (birth, death) {
+  if (!death) {
+    death = new Date().getFullYear();
+  }
+  return death - birth;
+}
```
MaoShizhong, line 6: Nit
```suggestion
};
```

12_findTheOldest/solution/findTheOldest-solution.js
```diff
@@ -9,11 +16,24 @@ const findTheOldest = function (array) {
...
+//};
+
+//const findTheOldest = function (people) {
+//  const peopleOldestToYoungest = people.toSorted(
+//    (person, nextPerson) => getAge(nextPerson) - getAge(person),
+//  );
+//
+//  const oldestPerson = peopleOldestToYoungest[0];
+//  return oldestPerson;
+//};
```
MaoShizhong, line 37: Another nit solely for the sake of readability. Lots of `//` really gets in the way :sweat_smile: 

```suggestion
/* ALTERNATIVE SOLUTION
const getAge = function (person) {
  // The nullish coalescing assignment operator
  // only does the assignment if the left side is "nullish" (evaluates to undefined or null)
  // if the left side has any other value, no assignment happens
  // here, we use ??= to set the current year for our subtraction below only if there is no year of death
  person.yearOfDeath ??= new Date().getFullYear();

  return person.yearOfDeath - person.yearOfBirth;
};

const findTheOldest = function (people) {
  const peopleOldestToYoungest = people.toSorted(
    (person, nextPerson) => getAge(nextPerson) - getAge(person),
  );

  const oldestPerson = peopleOldestToYoungest[0];
  return oldestPerson;
};
*/
```

12_findTheOldest/solution/findTheOldest-solution.js
nikitarevenco: Good point!

MaoShizhong (changes requested):

12_findTheOldest/solution/findTheOldest-solution.js
```diff
@@ -9,11 +16,26 @@ const findTheOldest = function (array) {
   });
 };
 
-const getAge = function (birth, death) {
-  if (!death) {
-    death = new Date().getFullYear();
-  }
-  return death - birth;
+// An alternative solution
```
MaoShizhong, line 19:
```suggestion
```
Oops, missed this

MaoShizhong (changes requested):

MaoShizhong (approved):
:rocket:

PR #460, opened 2024-05-03
feat(repeatString): add instruction to use loops for implementation rather than builtin method

nikitarevenco:
There is a builtin method for accomplishing the same outcome, so I have updated the instructions as the point of the exercise is to practice loops.

### DIFF
--- a/02_repeatString/README.md
+++ b/02_repeatString/README.md
@@ -19,8 +19,6 @@ You will notice in this exercise that there are multiple tests (see in file `rep
 
 - Take note of the above function call- how exactly is it being called?
 
-- You're going to want to use a loop for this one.
-
 - Create a variable to hold the string you're going to return, create a loop that repeats the given number of times and add the given string to the result on each loop.
 
 - If running `npm test repeatString.spec.js` returns results similar to the below, make sure you have enabled the rest of the tests, as described in the instructions above.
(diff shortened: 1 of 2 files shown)


MaoShizhong:
I'm hesitant on this one.

I believe the intention of the exercise is to get learners to get more practice with using loops. The hint quite firmly suggests this, but I feel for the exercise's intent, if loops are what it wants the learner to practise, it would be better placed in the actual exercise instructions instead of the hints section.

Basically saying that even though there is a built in string method for repeating strings, we want to practise loops so do it manually with loops instead. That way, they're also made aware of the string method.

nikitarevenco:
> I'm hesitant on this one.
> 
> I believe the intention of the exercise is to get learners to get more practice with using loops. The hint quite firmly suggests this, but I feel for the exercise's intent, if loops are what it wants the learner to practise, it would be better placed in the actual exercise instructions instead of the hints section.
> 
> Basically saying that even though there is a built in string method for repeating strings, we want to practise loops so do it manually with loops instead. That way, they're also made aware of the string method.

I've repurposed this PR

02_repeatString/README.md
MaoShizhong, line 1: Thanks for this. Mind removing the "You're going to want to use a loop for this one." hint as it's now redundant? The rest can be left as is.

MaoShizhong (changes requested):

02_repeatString/README.md
nikitarevenco: Ok, I removed it

MaoShizhong (approved):

PR #462, opened 2024-05-03
Exercise 8: Assign `power` to `Math.pow`

nikitarevenco:
I remember when I was at this stage I didn't ever think of something like this. For some reason I don't think many beginners will realise that this is possible to do with functions, so I think changing this little bit gives for a perfect example of how functions are passed by reference.

### DIFF
--- a/08_calculator/solution/calculator-solution.js
+++ b/08_calculator/solution/calculator-solution.js
@@ -14,9 +14,8 @@ const multiply = function (array) {
   return array.reduce((product, current) => product * current)
 };
 
-const power = function (a, b) {
-  return Math.pow(a, b);
-};
+// Functions in JavaScript are variables that be manipulated just like any other variable
+const power = Math.pow
 
 const factorial = function (n) {
   if (n === 0) return 1;
--- a/08_calculator/solution/calculator-solution.js
+++ b/08_calculator/solution/calculator-solution.js
@@ -14,7 +14,7 @@ const multiply = function (array) {
   return array.reduce((product, current) => product * current)
 };
 
-// Functions in JavaScript are variables that be manipulated just like any other variable
+// Functions in JavaScript are passed by reference
 const power = Math.pow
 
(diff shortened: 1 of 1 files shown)


MaoShizhong:
While I like the idea, I don't think this is an appropriate or necessary change to make.
The task, as with any other exercise and functions in the `calculator.js` file, are to complete the incomplete function in the file. Therefore, it's perfectly reasonable people won't even consider just assigning `const power = Math.pow;`, because that would require effectively going against instructions and removing the empty `function() { }`. Again, remember the point of the exercise is just to complete the tasks according to the spec however. I don't think changing (or even adding as an alternative) the solution to assigning `Math.pow` provides substantial benefit. It's perfectly sufficient to call `Math.pow` inside their own function. Some might even use `**` instead of `Math.pow`. Either is completely fine.

Functions passed as references will have been introduced in the Object basics lesson already, and is something people will come across plenty throughout the course. No need to provide what would seem a real left field solution for the sake of it.

nikitarevenco:
> While I like the idea, I don't think this is an appropriate or necessary change to make. The task, as with any other exercise and functions in the `calculator.js` file, are to complete the incomplete function in the file. Therefore, it's perfectly reasonable people won't even consider just assigning `const power = Math.pow;`, because that would require effectively going against instructions and removing the empty `function() { }`. Again, remember the point of the exercise is just to complete the tasks according to the spec however. I don't think changing (or even adding as an alternative) the solution to assigning `Math.pow` provides substantial benefit. It's perfectly sufficient to call `Math.pow` inside their own function. Some might even use `**` instead of `Math.pow`. Either is completely fine.
> 
> Functions passed as references will have been introduced in the Object basics lesson already, and is something people will come across plenty throughout the course. No need to provide what would seem a real left field solution for the sake of it.

Okay! Personally it took me a while to realise you can do manipulations like this with functions, despite knowing that they are passed by reference. But I do get where you are coming from!

PR #463, opened 2024-05-03
Exercise 5: Add solution that is more complicated

nikitarevenco:
In my world, the learner should be able to learn something interesting from all the exercises. In here I added a solution that is more advanced than what they would've been used to at this point. It uses `Array.from` with an object that has a `length` property 

And also uses this array only for its indexes to create an array of consecutive integers from min to max

### DIFF
--- a/05_sumAll/solution/sumAll-solution.js
+++ b/05_sumAll/solution/sumAll-solution.js
@@ -18,4 +18,16 @@ const sumAll = function (min, max) {
   return sum;
 };
 
+// A more advanced approach
+// const sumAll = function(min, max) {
+//  if (!Number.isInteger(min) || !Number.isInteger(max)) return "ERROR";
+//  if (min < 0 || max < 0) return "ERROR";
+//  create an array of consecutive integers from min to max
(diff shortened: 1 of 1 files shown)


MaoShizhong:
Thanks for making this suggestion @nikitarevenco.
I think when considering the purpose of this exercise, as well as the position within the curriculum, and the level of the typical learner at this point, I really can't see a lot of benefit in providing this sort of alternative solution. If anything, I believe it will be significantly more confusing for many learners when taking all of that into account.

Remember that at this point, many learners are not necessarily experienced enough to recognise what things are just "interesting" and "good practices worth investing time into understanding". We already have enough cases in the community where simpler, easier-to-read (and also very reasonable given the exposed concepts thus far) solutions are confusing enough to learners, and they stress over whether they need to devote more time into "mastering best practices". We wouldn't want to promote that, as the purpose of the exercises are to just pass them in whatever way according to the spec, using the things thought so far.

For that reason, this would be a strong push back from me, so I will be closing this particular issue.

PR #464, opened 2024-05-03
Exercise 5: Add solution using formula

nikitarevenco:
I feel like some learners will especially appreciate this solution. It's the fastest way to solve the problem and a good reminder that sometimes math is the answer to all questions and not some complex algorithm :')

### DIFF
--- a/05_sumAll/solution/sumAll-solution.js
+++ b/05_sumAll/solution/sumAll-solution.js
@@ -18,4 +18,8 @@ const sumAll = function (min, max) {
   return sum;
 };
 
+// A very mathematical approach. You'd probably be unlikely to stumble across this by chance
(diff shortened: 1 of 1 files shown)


MaoShizhong:
I'm going to push back on this specific solution for the purpose of this exercise because I believe our exercise solutions should demonstrate things that should be reasonable for most learners to read based on what they've learnt so far. I'm not bothered by demonstrating "efficient" or "clever" solutions if they aren't relevant to what we can expect most learners to been exposed to and reasonably figure out.

Just like with the palindrome exercise, it used to contain a regex solution. It's definitely what I would go for now but that isn't reasonable to show people at this point. More people were confused and even discouraged by seeing that than curious. Instead, providing a more verbose non-regex solution that only uses what learners have reasonably been exposed to in the curriculum, that's been received much better.

PR #478, opened 2024-06-04
Project Root: Add minimum npm & node version requirements

pixie-cheeks:
This change will prevent any old node or npm version from installing any packages. With this, npm will throw an error and exit immediately. This is to prevent the accidental use of unsupported versions which might cause some issues that are hard to pin-point.

## Because

A person on Discord was getting some syntax errors when running Jest after installing the packages. It took us a while to figure out that the person was using a really old version of node. The additions in this PR will prevent something like this from happening. If someone tries to run the `npm install` command with an outdated node/npm version inside the repo, then npm will exit with an error, informing the user their version is too low and what the minimum required version is.

## This PR

- Adds an engines field in the `package.json` file with the minimum supported node and npm versions
- Does the same as above in the `package-lock.json`
- Adds a `.npmrc` config file with an option that fails `npm install` with an error and exits (without this, npm only shows warnings)

## Additional Information

- [This](https://discord.com/channels/505093832157691914/1179952911548174346/1246911497767096412) was the discussion mentioned in the because section.
- Link to the SO answer that enabled this change: https://stackoverflow.com/a/29349773/20913895
- Method used to find the minimum version requirements for the repo: https://stackoverflow.com/a/70950326/20913895
- Finally, all the test files run properly in the minimum required versions

## Pull Request Requirements

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,10 @@
         "jest": "^29.6.4",
         "jest-cli": "^29.6.4",
         "plop": "^4.0.1"
+      },
+      "engines": {
+        "node": ">=18.0.0",
+        "npm": ">=8.5.5"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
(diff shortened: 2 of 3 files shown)


wise-king-sullyman (approved):
Sounds like a great change, thanks!

PR #505, opened 2024-11-13
02_repeatString - Line-32, replace "odin" with "hey"

bhagyeshsp:
The original code had "hey" as the seed string value and hence the comment on line-29 mentions "hey". But the last commit on the code changed "hey" to "odin".

We can either change instances of comment or the code. I propose changing the code instance from "odin" to "hey" for consistency.

## Because

There is an inconsistency between the comment on line-29 and its corresponding code on line-32. It provides clarity to the learner by maintaining the same example of "hey" instead of "odin".

## This PR

- Replaced two instances of "odin" with "hey" on line-32

## Issue

Closes #XXXXX

## Pull Request Requirements

-   [ ] If this PR includes any changes that affect the solution of an exercise, I've also updated the solution in the `/solutions` folder

### DIFF
--- a/02_repeatString/repeatString.spec.js
+++ b/02_repeatString/repeatString.spec.js
@@ -29,7 +29,7 @@ describe('repeatString', () => {
     /*The .match(/((hey))/g).length is a regex that will count the number of heys
     in the result, which if your function works correctly will equal the number that
     was randomly generated. */
-    expect(repeatString('odin', number).match(/((odin))/g).length).toEqual(number);
+    expect(repeatString('hey', number).match(/((hey))/g).length).toEqual(number);
   });
   test.skip('works with blank strings', () => {
     expect(repeatString('', 10)).toEqual('');


MaoShizhong:
In #460 the instructions were made clear that `String.prototype.repeat` should not be used and the implementation should be done manually via loops. I wonder if the team would agree that on top of fixing the `'hey'` thing, the test be changed to 
```js
expect(repeatString('hey', number)).toBe('hey'.repeat(number));
```

Then the big regex explanation comment can be removed.

bhagyeshsp:
> In #460 the instructions were made clear that `String.prototype.repeat` should not be used and the implementation should be done manually via loops. 

Thanks, I saw #460  I'm unsure how it is related to this PR.
The current PR is more content-specific than code-specific.

> I wonder if the team would agree that on top of fixing the `'hey'` thing, the test be changed to
> 
> ```js
> expect(repeatString('hey', number)).toBe('hey'.repeat(number));
> ```
> 
The test has been already changed to its present form since August, 2024 under #460 
This PR proposes getting the comment, code and example in alignment by maintaining "hey" throughout.


> Then the big regex explanation comment can be removed.

I found the regex explanation educational.

MaoShizhong:
My comment was more an addition to yours and directed at the team when they review as opposed to directed at you, sorry if that wasn't clear. Regex was previously removed from comments and solutions in some of these exercises due to it being rather out of scope for this part of the curriculum. Hence the suggestion to the team that while we're at it, it may be best to apply the same approach here.

JoshDevHub (changes requested):
@bhagyeshsp This seems like a good change to me. Sorry it took us a while to look at it.

I have one change to request: the [solutions test](https://github.com/TheOdinProject/javascript-exercises/blob/main/02_repeatString/solution/repeatString-solution.spec.js) needs to make this same change. Would you be able to add a commit that changes it there as well?

JoshDevHub:
@MaoShizhong I also think what you've said on this makes sense. Feel free to make a PR (or an issue if you don't want to work on it) that removes regex and uses `String.prototype.repeat`

bhagyeshsp:
> @bhagyeshsp This seems like a good change to me. Sorry it took us a while to look at it.
> 
> I have one change to request: the [solutions test](https://github.com/TheOdinProject/javascript-exercises/blob/main/02_repeatString/solution/repeatString-solution.spec.js) needs to make this same change. Would you be able to add a commit that changes it there as well?

Hi @JoshDevHub, implemented the commit in the requested solution too.

JoshDevHub (approved):
You could have just added the commit to this PR instead of opening a new one, but that's okay. Something for you to keep in mind in the future :thumbsup:

bhagyeshsp:
> You could have just added the commit to this PR instead of opening a new one, but that's okay. Something for you to keep in mind in the future 👍

Oh, I will keep that in mind in the future. Still learning the nitty-gritty.
Thanks for considering the merge.

PR #523, opened 2025-02-04
02_repeatString: Add test preventing use of `String.prototype.repeat`

MaoShizhong:
## Because

The instructions for 02_repeatString make it clear that the exercise is for practising loops, so the user must not use the built-in `String.prototype.repeat` method to complete this, even if it leads to the same result. Despite this, the test suite doesn't actually prevent its usage. It would be worth adding a single test that fails upon its use and explains why, in order to encourage practising loops.

## This PR

- Adds test preventing use of `String.prototype.repeat` in both main and solution test files

## Issue

N/A

## Pull Request Requirements


JoshDevHub (approved):
:rocket:

PR #550, opened 2025-06-01
07_reverseString: handle Unicode surrogate pairs in reverseString solution 

eyad-alkhalidy:
## Because

The original `reverseString` solution used `.split("")`, which does not handle complex Unicode characters correctly. This can lead to incorrect behavior when the input includes surrogate pairs (e.g., emojis or some accented characters). This PR modifies the implementation to avoid that issue and includes a comment to make learners aware of the underlying limitation.

## This PR

-   Updated the `reverseString` solution to use `Array.from()` instead of `.split("")`
-   Added a comment explaining the issue with surrogate pairs and Unicode handling in JavaScript

## Issue

N/A

## Additional Information

We've already been exposed to this problem in the MDN Reference page for String, linked [here](https://www.theodinproject.com/lessons/foundations-data-types-and-conditionals#:~:text=MDN%20documentation%20for%20strings). This is why I think we can consider it in this exercise.

## Pull Request Requirements

### DIFF
--- a/07_reverseString/solution/reverseString-solution.js
+++ b/07_reverseString/solution/reverseString-solution.js
@@ -1,5 +1,16 @@
 const reverseString = function (string) {
-  return string.split("").reverse().join("");
+  return Array.from(string).reverse().join("");
+  
+  /*
+  The following statement: `return string.split("").reverse().join("");`
+  also works but only if the passed string doesn't include any complex Unicode character,
+  i.e. a characher stored in a surrogate pair, which is a pair of 16-bit code units that represents a single character.
+  Example: emojis as in "Great 👍" and some accented characters as in "mañana".
+  
+  The `split()` method splits the String by UTF-16 code units and will separate surrogate pairs resulting in invalid characters.
+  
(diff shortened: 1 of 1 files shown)


mao-sz:
Thanks for the proposal - discussed this with the team and we've decided to keep `.split()`.

The angle behind this is that there needs to be a significant problem to warrant a change, to avoid "sideways" changes just for the sake of it. The problem of surrogate pairs is both beyond the scope we'd want for this exercise, as well as incomplete due to `Array.from` having the same issues with stuff like grapheme clusters, and that's even further out of scope.

While `Array.from()` or spread syntax are both newer, they don't make `.split()` outdated by any means. You will still find `.split()` used for this purpose in many codebases, old and new, as well as when researching about turning a string into an array of chars. AFAIK there is no official recommendation that `Array.from()` be used over `.split()` in general, nor a major shift in conventions.

Since we don't want to expand the scope of this exercise, purposely changing the solution to `Array.from()` may raise questions as to why it was made, and may confuse learners who think it's commentary that `.split()` should not be used (which it won't be).

PR #574, opened 2025-08-28
`15_getTheTitles` : Added more tests to avoid

maitreverge:
## Because
Despite the `README.md` mentionning the mandatory use of a `a built-in javascript method`, we can still pass this exercice with a simple solution including a bracket notation like:

```js
const getTheTitles = function(arr) {
	let result = []
	
	result.push(arr[0]["title"]);
	result.push(arr[1]["title"]);
	return result;
};

// Do not edit below this line
module.exports = getTheTitles;
```

I added more tests to avoid this simple solution.

## This PR

- The file `getTheTitles.spec.js` and `getTheTitles-solution.spec.js` has been changed.
- The following variables has been added...

```js
  // An array dynamically generated
  const largeBooks = Array.from({ length: 1000 }, (_, i) => ({
    title: `Title${i}`,
    author: `Author${i}`,
  }));

  // And an array with different values types as book title.
  const booksWithNonStringTitles = [
    {
		title: 123,
		author: 'A',
	},
    {
		title: null,
		author: 'B',
	},
    {
		title: 'Valid',
		author: 'C',
	},
  ];
```

- ...and their related test

```js
  test.skip('returns empty array for empty input', () => {
    expect(getTheTitles([])).toEqual([]);
  });
  test.skip('handles large array', () => {
    expect(getTheTitles(largeBooks)).toEqual(largeBooks.map((b) => b.title));
  });
  test.skip('handles non-string titles', () => {
    expect(getTheTitles(booksWithNonStringTitles)).toEqual([
      123,
      null,
      'Valid',
    ]);
  });
```

## Pull Request Requirements

-   [ x ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
-   [ x ] The title of this PR follows the `location of change: brief description of change` format, e.g. `01_helloWorld: Update test cases`
-   [ x ] The `Because` section summarizes the reason for this PR
-   [ x ] The `This PR` section has a bullet point list describing the changes in this PR
-   [ x ] If this PR addresses an open issue, it is linked in the `Issue` section
-   [ x ] If this PR includes any changes that affect the solution of an exercise, I've also updated the solution in the `/solutions` folder

### DIFF
--- a/15_getTheTitles/getTheTitles.spec.js
+++ b/15_getTheTitles/getTheTitles.spec.js
@@ -1,18 +1,51 @@
 const getTheTitles = require('./getTheTitles')
 
 describe('getTheTitles', () => {
-    const books = [
-      {
-        title: 'Book',
-        author: 'Name'
-      },
-      {
-        title: 'Book2',
-        author: 'Name2'
-      }
-    ]
+  const books = [
+    {
+      title: 'Book',
+      author: 'Name'
+    },
+    {
+      title: 'Book2',
+      author: 'Name2'
+    },
+  ];
+
+  const largeBooks = Array.from({ length: 1000 }, (_, i) => ({
+    title: `Title${i}`,
+    author: `Author${i}`,
+  }));
+
+  const booksWithNonStringTitles = [
+    {
+		title: 123,
+		author: 'A',
+	},
+    {
+		title: null,
+		author: 'B',
+	},
+    {
+		title: 'Valid',
+		author: 'C',
+	},
+  ];
 
   test('gets titles', () => {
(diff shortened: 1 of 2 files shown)


mao-sz (changes requested):
Thanks @maitreverge 

I'm on board with the `booksWithNonStringTitles` extra case, mainly because it has a different length.
I'd like the `largeBooks` array and test case removed though, because it actually leaks a solution in the test assertion :P

Technically, if someone loops over the array and `.push`es to a new array, they're using a built-in JS method so that's perfectly acceptable, and the exercise is not trying to enforce the use of `.map` only.

PR #577, opened 2025-09-04
Migrate CircleCI workflows to GitHub actions

mao-sz:
## Because

There is a CircleCI config set up but this doesn't seem to actually be used for any PRs affecting relevant files. In a similar vein to https://github.com/TheOdinProject/odin-bot-v2/pull/648, it'd be easier to make use of GH actions since it's integrated and we get unlimited action credits as well.

We only really need the workflow to run on changes to files in the `solution` dirs, and only run solution tests (since exercise tests suites will likely fail and skip many tests).

### Caveats

- Runs all solution tests, not just for the changed solution dirs (seemed a little unnecessarily complex to configure, and running all solution tests is a negligible impact anyway).
- Won't check to keep tests for exercise and solution in sync - got to do this manually. Can't think of a reasonable workflow that can ensure this (has many caveats of its own).

## This PR

- Adds GH workflow to run solution tests on changes to any solution dir
- Removes unused CircleCI config

## Issue

N/A

## Pull Request Requirements


KevinMulhern (approved):
Looks great to me 🚀

PR #660, opened 2026-07-25
14_fibonacci: add negative number example to readme

GPerencine:
## Because
The README for the Fibonacci exercise does not mention that passing a negative number should return "OOPS", even though this behavior is tested in `fibonacci.spec.js`. Adding this example clarifies the exercise requirements.

## This PR
- Adds `fibonacci(-1); // returns "OOPS"` example to the README code block.

## Issue
N/A

## Additional Information
N/A

## Pull Request Requirements
- [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `01_helloWorld: Update test cases`
- [x] The `Because` section summarizes the reason for this PR
- [x] The `This PR` section has a bullet point list describing the changes in this PR
- [ ] If this PR addresses an open issue, it is linked in the `Issue` section
- [ ] If this PR includes any changes that affect the solution of an exercise, I've also updated the solution in the `/solutions` folder


mao-sz (changes requested):
Instead of this, we should just add onto the end of the README that it should return `"OOPS"` if passed a negative number. The example code block is only for clarifying the previous instruction about the sequence starting at 1 instead of 0. Just having text instruction for negatives is sufficient.

GPerencine:
Thanks for the feedback! I've updated the README to remove the code example and only added the text instruction at the end as requested.

mao-sz (changes requested):
Since the code block is specifically to give an example for the sentence about the Fibonacci sequence, the instruction about the negative arg should go after it at the end, not between it and its associated instruction.

GPerencine:
Thanks for the feedback again! I've moved the note regarding negative numbers to the end of the README, right after the code example.

foundations/14_fibonacci/README.md
```diff
@@ -8,3 +8,4 @@ In this exercise, the Fibonacci sequence used is 1, 1, 2, 3, 5, 8, etc. (i.e. st
 fibonacci(4); // returns the 4th member of the series: 3  (1, 1, 2, 3)
 fibonacci(6); // returns 8
 ```
+Note: if given a negative number, the function should return "OOPS".
```
mao-sz: "Note" feels out of place here. It's not really a side note, it's just a normal part of the instructions. Also added a blank line and backticks for markdown formatting purposes.
```suggestion

If given a negative number, the function should return `"OOPS"`.
```

mao-sz (changes requested):

foundations/14_fibonacci/README.md
GPerencine: Thanks for the refinement! I've committed the suggested change.

mao-sz (approved):

PR #677, opened 2026-09-13
Foundations Fibonacci: Streamline expected error-handling behavior

Vairiation:
## Because

Previous logic for solution allows fibonacci('hello') to return 1 instead of 'OOPS'. While this is not a breaking change, I felt as though it was relevant when looking at the conditional logic.

## This PR

- Add `!count` to if conditional block that returns `"OOPS"` so it now reads `if (!count || count < 0) return "OOPS"`
- Move `if (count == 0) return 0` before `if (!count || count < 0) return "OOPS"` to allow proper handling of `count = 0`

## Issue

Closes #XXXXX

## Pull Request Requirements

-   [ x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
-   [ x] The title of this PR follows the `location of change: brief description of change` format, e.g. `01_helloWorld: Update test cases`
-   [ x] The `Because` section summarizes the reason for this PR
-   [ x] The `This PR` section has a bullet point list describing the changes in this PR
-   [ x] If this PR addresses an open issue, it is linked in the `Issue` section
-   [ x] If this PR includes any changes that affect the solution of an exercise, I've also updated the solution in the `/solutions` folder


mao-sz (changes requested):
Thanks for suggesting this. I'm not actually convinced by it though. The point of the exercise isn't to make a robust function for real world usage that covers most/all possible cases, but rather to just practise some fundamentals in order to pass all the tests, for which `NaN` handling isn't.

That being said, looking at the current tests, I'm not a fan of the last few that deal with strings. Not really seeing the point having them in this exercise specifically; there are plenty of opportunities for parsing/converting strings to numbers throughout the whole curriculum.

---

I'd like to propose a different and slightly larger change to this exercise. If you're happy to do it, then just do the work in this branch. Otherwise, you can close the PR and I'll get the work handled another way.

I think instead of testing for string->number conversion, we should amend the instructions and tests so that if the argument is a negative number, `NaN` or not a number-type at all, return `"OOPS"`, i.e. only proceed with positive numbers.

Means it focuses more on doing the fibonacci stuff and just shoving aside all other types of values, no other type-specific behaviour.

Vairiation:
Hey mao-sz, I just made the requested changes to the solution function, solution test suite, and student test suite.

I added in a check for NaN specifically since its type is a number and could slip through checking `typeof countArg !== 'number'` alone.

Let me know if that was the direction you were looking for. Also, this is my first pull request on a non-individual repo so let me know if I did something incorrectly.

foundations/object_basics/03_fibonacci/solution/fibonacci-solution.js
```diff
@@ -2,14 +2,11 @@ const fibonacci = function(countArg) {
...
-        count = countArg
-    }
-
-    if (count < 0) return "OOPS";
-    if (count == 0) return 0;
+    if (typeof countArg !== 'number' || countArg < 0 || Number.isNaN(countArg)) {
+        return "OOPS";
+    } else count = countArg;
+    
+    if (count === 0) return 0;
```
mao-sz: No need for separate `count` and `countArg` anymore, we can just use the param and call it `count` for simplicity. For some reason, GitHub won't let me include lines 1 and 4 in the suggestion below so you'll have to manually rename the param and remove line 4.
```suggestion
    if (typeof count !== 'number' || count < 0 || Number.isNaN(count)) {
        return "OOPS";
    }    
    if (count === 0) {
        return 0;
    }
```

foundations/object_basics/03_fibonacci/solution/fibonacci-solution.js
```diff
@@ -24,6 +21,8 @@ const fibonacci = function(countArg) {
 
 };
 
+console.log(fibonacci('hello'))
+
```
mao-sz: This needs to go.
```suggestion
```

foundations/object_basics/03_fibonacci/solution/fibonacci-solution.js
```diff
@@ -32,3 +31,4 @@ const fibonacci = function(countArg) {
 // return fib[count];
 
 module.exports = fibonacci;
+
```
mao-sz: Same here
```suggestion
```

foundations/object_basics/03_fibonacci/solution/fibonacci-solution.js
```diff
@@ -24,6 +21,8 @@ const fibonacci = function(countArg) {
```
mao-sz: While you're at it, might as well remove this rogue blank line too.
```suggestion
```

foundations/object_basics/03_fibonacci/fibonacci.spec.js
```diff
@@ -22,16 +22,28 @@ describe('fibonacci', () => {
...
+  test.skip('doesn\'t accept Boolean', () => {
+    expect(fibonacci(false)).toBe("OOPS");
+  });
+  test.skip('doesn\'t accept Array', () => {
+    expect(fibonacci([1])).toBe("OOPS");
+  });
+  test.skip('doesn\'t accept Object', () => {
+    expect(fibonacci({num: 1})).toBe("OOPS");
+  });
+})
```
mao-sz: I don't think we need individual tests for each data type. This is only a small side part of the exercise so it'll suffice to have a test for negatives, NaN and "non-numbers".
```suggestion
  test.skip('doesn\'t accept NaN', () => {
    expect(fibonacci(NaN)).toBe("OOPS");
  });
  test.skip('doesn\'t accept non-numbers types', () => {
    expect(fibonacci("0")).toBe("OOPS");
    expect(fibonacci([])).toBe("OOPS");
    expect(fibonacci({})).toBe("OOPS");
  });
})
```

foundations/object_basics/03_fibonacci/README.md
```diff
@@ -2,11 +2,12 @@
 
 Create a function that returns a specific member of the Fibonacci sequence (series of numbers in which each number is the sum of the two preceding numbers). To learn more about Fibonacci sequences, go to: https://en.wikipedia.org/wiki/Fibonacci_sequence
 
-In this exercise, the Fibonacci sequence used is 1, 1, 2, 3, 5, 8, etc. (i.e. starting at 1, not 0).
+In this exercise, the Fibonacci sequence used is 1, 1, 2, 3, 5, 8, etc. (i.e. starting at 1, not 0). The function should only accept number inputs and reject anything else.
```
mao-sz: No need to duplicate the instruction - it's already mentioned later.
```suggestion
In this exercise, the Fibonacci sequence used is 1, 1, 2, 3, 5, 8, etc. (i.e. starting at 1, not 0):
```

foundations/object_basics/03_fibonacci/README.md
```diff
@@ -2,11 +2,12 @@
...
 
 ```javascript
 fibonacci(4); // returns the 4th member of the series: 3  (1, 1, 2, 3)
 fibonacci(6); // returns 8
+fibonacci("6"); // returns "OOPS"
```
mao-sz: We don't need this example here. This code block isn't an example of the entire instructions, it's an example of the previous point only (about sequence this exercise uses beginning at 1 and not 0).
```suggestion
```

foundations/object_basics/03_fibonacci/README.md
```diff
@@ -2,11 +2,12 @@
...
 ```
 
-If given a negative number, the function should return `"OOPS"`.
\ No newline at end of file
+If given a negative number, `NaN`, or not a number type at all, the function should return `"OOPS"`, i.e. only proceed with positive numbers..
```
mao-sz:
```suggestion
If given a negative number, `NaN`, or not a number type at all, the function should return `"OOPS"`, i.e. only proceed with positive numbers.
```

mao-sz (changes requested):

foundations/object_basics/03_fibonacci/solution/fibonacci-solution.js
Vairiation: Agreed, the countArg seemed redundant but wasn't sure if you wanted to keep it in. That is weird that github wouldn't let you include lines 1-4. Should be resolved now

foundations/object_basics/03_fibonacci/fibonacci.spec.js
Vairiation: Interesting, I didn't know that you could test multiple function calls with one test.

Vairiation:
I think that should be everything from your suggestions. Let me know if I missed anything!

mao-sz (changes requested):
The test changes need to be duplicated across the solution test file as well

mao-sz (approved):
:rocket:

